PR #1, opened 2018-08-28
add AgateDecorator and supporting files

sugardave:
Based on Zircon, so some of it may be slightly out of date.  But, it's close to pristine and ready for customization.


sugardave:
Womp, forgot to exclude IDE files, canceling and remaking.

PR #22, opened 2018-09-04
Feature/IconItem

teckliew:
`IconItem`
screen shot 2018-09-04 at 1 07 49 pm
screen shot 2018-09-04 at 1 07 11 pm

Also updated `Item` and `LabeledItem` styles.


Djspaceg:
I spent a little time trying to get this to handle long content better, but I didn't have much luck. I'm merging this anyway, but if someone comes up with a better way to handle the "long content" case, I'd be happy to see it.

PR #35, opened 2018-09-07
Feature/ToggleButton

teckliew:
Add `ToggleButton`.
Refer to https://github.com/enactjs/agate-apps/pull/16 for Storybook sample.

There's two versions of `ToggleButton`: one with the `toggleIndicator` and one without.
`toggleIndicator` is the underline below the text/icon. Set `toggleIndicator` prop to true to show the underline, otherwise it'll be a normal button with `selected` styling.

togglebutton-on-indicator2
togglebutton-off-indicator2
togglebutton-on2
togglebutton-off2
togglebutton-on
togglebutton-off
togglebutton-on-indicator
togglebutton-off-indicator


ToggleButton/ToggleButton.js
```diff
@@ -0,0 +1,212 @@
+/**
+ * Agate styled button components and behaviors.
+ *
+ * @example
+ * <Button small>Hello Enact!</Button>
+ *
+ * @module agate/Button
+ * @exports Button
+ * @exports ButtonBase
+ * @exports ButtonDecorator
```
germboy: While it may still be too early to have proper documentation (descriptions) for agate components, I'd like to make sure we are providing proper examples and @export docs. Could we update the `Button` references to `ToggleButton` and make sure we're only including what we're exporting here (`ToggleButton`, `ToggleButtonBase`)?

ToggleButton/ToggleButton.js
```diff
@@ -0,0 +1,212 @@
...
+ */
+
+import kind from '@enact/core/kind';
+import Pure from '@enact/ui/internal/Pure';
+import Toggleable from '@enact/ui/Toggleable';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import Button from '../Button';
+// import {MarqueeDecorator} from '../Marquee';
```
germboy: I'm guessing this is for a future implementation, but it looks like `agate/Button` has the same plans. If it's implemented in `Button`, we could probably remove this. Thoughts?

ToggleButton/ToggleButton.js
```diff
@@ -0,0 +1,212 @@
...
+	propTypes: /** @lends agate/Button.ButtonBase.prototype */ {
+		/**
+		 * The background-color opacity of this button.
+		 *
+		 * * Values: `'translucent'`, `'lightTranslucent'`, `'transparent'`
+		 *
+		 * @type {String}
+		 * @public
+		 */
+		backgroundOpacity: PropTypes.oneOf(['translucent', 'lightTranslucent', 'transparent']),
```
germboy: Do we have styles defined for these values? I don't see them or notice visual changes when I change these values. It appears these would be passed to `Button` but it's not clear to me that they're handled there either.

ToggleButton/ToggleButton.js
```diff
@@ -0,0 +1,212 @@
...
+		 * Enforces a minimum width on the Button.
+		 *
+		 * *NOTE*: This property's default is `true` and must be explicitly set to `false` to allow
+		 * the button to shrink to fit its contents.
+		 *
+		 * @type {Boolean}
+		 * @default true
+		 * @public
+		 */
+		minWidth: PropTypes.bool,
```
germboy: I notice that when we toggle between labels of varying length ("off" and "on"), the width of the `ToggleButton` will grow/shrink somewhat, which seems unexpected when using `minWidth`. Would this prop stabilize that behavior? If not, we may want to look into adjusting that specifically.

ToggleButton/ToggleButton.js
```diff
@@ -0,0 +1,212 @@
...
+
+		/**
+		 * Indicates the button is 'on'.
+		 *
+		 * @type {Boolean}
+		 * @default false
+		 * @public
+		 */
+		selected: PropTypes.bool,
```
germboy: This appears inactive since it's passed to `Button`, where it's also inactive there. Is that correct?

ToggleButton/ToggleButton.less
```diff
@@ -0,0 +1,84 @@
+// ToggleButton.less
+//
+@import "../styles/mixins.less";
+@import "../styles/fonts.less";
+@import "../styles/skin.less";
+
+.toggleButton {
+	.applySkins({
+		// .padding-start-end(@agate-button-h-padding, @agate-toggle-button-text-switch-gutter + @agate-button-h-padding + @agate-toggle-button-switch-width + @agate-toggle-button-switch-border-width);
```
germboy: Can we remove this, if it's indeed cruft?

styles/colors-titanium.less
```diff
@@ -101,4 +106,19 @@
...
+// ---------------------------------------
+@agate-toggle-border-color: @agate-gray;
+@agate-toggle-indicator-border-color: transparent;
+@agate-toggle-focus-text-color: @agate-foreground-focus;
+@agate-toggle-button-switch-on-bg-color: @agate-highlight;
+@agate-toggle-button-switch-on-border-color: @agate-highlight;
+@agate-toggle-button-switch-on-focus-bg-color: @agate-white;
+@agate-toggle-button-switch-on-focus-border-color: @agate-white;
+@agate-toggle-button-switch-off-bg-color: @agate-gray;
+@agate-toggle-button-switch-off-border-color: @agate-white;
```
germboy: I don't see this used anywhere. Can we remove it?

ToggleButton/ToggleButton.js
teckliew: removed!

ToggleButton/ToggleButton.js
teckliew: fixed!

ToggleButton/ToggleButton.js
teckliew: woops we don't need this for ToggleButton

ToggleButton/ToggleButton.js
teckliew: I'll remove this for now, and have leave the min-width up to the consumer

ToggleButton/ToggleButton.js
teckliew: It is passed down to Button, but it's being used for accessibility and styles

ToggleButton/ToggleButton.less
teckliew: yep

styles/colors-titanium.less
teckliew: done!

PR #38, opened 2018-09-07
Feature/picker

rundmt:
Made a basic picker to sort of look like AGL, but added three slots instead of two.  Maybe we should just switch out to icon buttons looking for some feedback.

### DIFF
--- a/Picker/Picker.less
+++ b/Picker/Picker.less
@@ -11,21 +11,25 @@
 	justify-content: center;
 	width: 70px;
 	font-weight: 500;
-	font-size: 30px;
+	font-size: 24px;
 
-	.item{
+	.item {
 		color: inherit;
 		width: 100%;
 		line-height: 40px;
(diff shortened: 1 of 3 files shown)


Picker/Picker.less
```diff
@@ -0,0 +1,34 @@
...
+	justify-content: center;
+	width: 70px;
+	font-weight: 500;
+	font-size: 24px;
+
+	.item {
+		color: inherit;
+		width: 100%;
+		text-align: center;
+		&.currentIndex{
```
germboy: minor nit: Can we add a space between `currentIndex` and the bracket?

rundmt:
@germboy Thanks for the feedback. I just fixed up the Spotlight stuff and some of the nitpicks. I don't think we want to add a scrolling like behavior right now just because of time. But eventually we can add something like that in.

PR #50, opened 2018-09-14
add state to TabbedPanels

sugardave:
`selected` prop from the `Group` is synced to the `index` prop for the `Panels`.


sugardave:
When this is merged, should also merge https://github.com/enactjs/agate-apps/pull/47

PR #56, opened 2018-09-21
Added ColorPicker

Djspaceg:
Choose any color you like! The color comes back as `value`, like any other input.

### DIFF
--- a/ColorPicker/ColorPicker.js
+++ b/ColorPicker/ColorPicker.js
@@ -1,0 +1,170 @@
+/**
+ * Agate styled labeled divider components and behaviors
+ *
+ * @example
+ * <ColorPicker
+ *   spacing="medium"
+ * >
+ *   A group of related components
+ * </ColorPicker>
+ *
+ * @module agate/ColorPicker
+ * @exports ColorPicker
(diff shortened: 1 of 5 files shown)


ColorPicker/ColorPicker.js
```diff
@@ -0,0 +1,112 @@
+/**
+ * Agate component to allow the user to choose a color.
+ *
+ * @example
+ * <ColorPicker
+ *   spacing="medium"
+ * >
+ *   A group of related components
```
teckliew: It seems like `ColorPicker` doesn't handle children, maybe update the doc to illustrate `ColorPicker` is used as a self-closing component.

Djspaceg: The docs aren't updated to reflect the component. I was hoping to save time by getting feedback on the functionality and approach before filling in all of the documentation. I'll update the docs now though.

teckliew: Pretty much the only feedback I have because I was trying to figure out how to use it at a glance, otherwise it looks good!

Djspaceg:
`ColorPicker` has been updated with real documentation.

PR #70, opened 2018-10-24
Agate Progressbar

teckliew:
Added a simple progress bar for Agate. To be customized more later.
Sampler for ProgressBar here https://github.com/enactjs/agate-apps/pull/80

carb
titan
elc


styles/variables-carbon.less
```diff
@@ -144,6 +144,13 @@
 @agate-popup-closebutton-position-left: auto;
 @agate-popup-closebutton-position-right: 0;
 
+// ProgressBar
+// ---------------------------------------
+@agate-progressbar-height: 10px;
```
Djspaceg: Multiple of 3.

styles/variables-electro.less
```diff
@@ -149,6 +149,13 @@
 @agate-popup-closebutton-position-left: auto;
 @agate-popup-closebutton-position-right: 18px;
 
+// ProgressBar
+// ---------------------------------------
+@agate-progressbar-height: 10px;
```
Djspaceg: How do you feel about making Electro and Titanium taller (and different heights)? Titanium could be the same height as the Slider while Electro could be somewhere in the middle.

Djspaceg: Also, if small progress bars are a design requirement for the dashboard, maybe it's time to add a `small` prop to progress bar, which applies the `.small` class to get a shorter height option?

Issue #67, opened 2018-10-22
Color Picker Palette Popup


Djspaceg:
We need a color picker popup UI for the color picker component when the system is unable to provide a UI. This will likely replace the built-in OS popup that appears on macOS.

## Requirements:
* The popup should appear when a user activates a color picker and remain open until they close it, or is dismissed by clicking somewhere else
* The selection method should allow the full range of colors from everything between #000000 and #ffffff
* The output should be in the hex (web-safe) color format (ex: `#rrggbb`)
* The interface should:
   * visually represent at least HSL (hue, saturation, lightness), and possibly other methods.
   * work with both touch and pointer.
   * use "math" to calculate the value, rather than reading rasterized pixels from a canvas element.
   * be dynamically scalable (use linear gradients) and possibly filters as blend-modes
* The input value should be "pre-selected" (represented in the UI) when popup opens.
* Color swatches (preset colors) definable in an array of hex strings as a prop on the picker popup component. Clicking/activating one should immediately update the popup's color selection, and not dismiss the popup (allowing the user to tune that color, not the preset, just the output)
* Use the module `color-convert` for color conversion. It's already part of Agate

## Nice to have
* 2 different styles of color seleciton, i.e. a 2D vertical/horizontal color space representation plus a lightness slider; and a set of 3 sliders for hue, saturation, and lightness
* at least two color picking mechanisms: HSL, HSV, direct HEX input, etc

Here's a link to some HSL color theory:
https://en.wikipedia.org/wiki/HSL_and_HSV

_**Example of a 2D picker with one slider:**_
<img width="245" alt="screen shot 2018-10-22 at 4 54 38 pm" src="https://user-images.githubusercontent.com/386529/47325949-57c83600-d61b-11e8-8039-a30c4c6a45c2.png">

_**Example of a 3-slider picker:**_
<img width="248" alt="screen shot 2018-10-19 at 1 37 28 pm" src="https://user-images.githubusercontent.com/386529/47325941-526aeb80-d61b-11e8-9b84-92ad4dc8afc6.png">

_**Note:** The color on the both of the example's secondary sliders change based on the currently selected hue._

teckliew:
Since we are simplifying this color picker to having limited number of color swatches, I'm thinking about making the color picker expandable instead of a popup.
The expandable color picker includes a swatch drawer that slides out and overlays the color picker's neighbors (to avoid layout repaint).
[...]

PR #72, opened 2018-11-02
Feature/expandable color picker

teckliew:
ColorPicker with an expandable drawer.
Usage example:
```
<ColorPicker defaultValue="#ffcc00" onChange={handleChange}>
  {['#fff', '#999', '#000', 'red', 'blue', 'green']}
</ColorPicker>
```

Consideration:
It is possible to implement the drawer like `Tooltip` for better positioning configurations, but it adds another level of complexity so it can be added later when needed. For now, `swatchDrawer` can be customized through `css` prop.

colopickerdemo
Resolves https://github.com/enactjs/agate/issues/67

### DIFF
--- a/ColorPicker/ColorPicker.less
+++ b/ColorPicker/ColorPicker.less
@@ -1,4 +1,4 @@
-// ExpandableColorPicker.less
+// ColorPicker.less
 //;
 @import "../styles/mixins.less";
 @import "../styles/skin.less";
@@ -21,7 +21,7 @@
 
 	.applySkins({
 		.swatchDrawer {
-			background-color: rgba(0, 0, 0, 0.5);
+			background-color: @agate-colorpicker-bg-color;
 			border-radius: @agate-colorpicker-swatch-border-radius;
 		}
 	});
--- a/ColorPicker/ColorPicker.less
+++ b/ColorPicker/ColorPicker.less
@@ -19,10 +19,14 @@
 		}
 	}
 
+	.group {
+		display: block;
+	}
+
 	.applySkins({
 		.swatchDrawer {
 			background-color: @agate-colorpicker-bg-color;
-			border-radius: @agate-colorpicker-swatch-border-radius;
+			border-radius: @agate-colorpicker-swatchdrawer-border-radius;
 		}
 	});
 }
--- a/ColorPicker/ColorPicker.less
+++ b/ColorPicker/ColorPicker.less
@@ -1,37 +1,28 @@
-// Divider.less
+// ExpandableColorPicker.less
 //;
 @import "../styles/mixins.less";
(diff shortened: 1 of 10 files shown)


ColorPicker/ColorPicker.less
```diff
@@ -1,37 +1,28 @@
-// Divider.less
+// ExpandableColorPicker.less
```
Djspaceg: "ColorPicker.less" ;)

ColorPicker/ColorPicker.less
```diff
@@ -1,37 +1,28 @@
...
-		}
-
-		.colorInput {
-			padding: 0;
-			width: 100%;
-			height: 100%;
-			opacity: 0;
+	.applySkins({
+		.swatchDrawer {
+			background-color: rgba(0, 0, 0, 0.5);
```
Djspaceg: This should be customizable with a skin prop.

ColorPicker/SwatchButton.less
```diff
@@ -0,0 +1,32 @@
+// SwatchButton.less
+//;
+@import "../styles/mixins.less";
+@import "../styles/skin.less";
+
+.swatchButton {
+	.applySkins({
+		margin: 10px;
```
Djspaceg: Not a multiple of 3 and should be a skin variable.

ColorPicker/SwatchButton.less
```diff
@@ -0,0 +1,32 @@
...
+		}
+
+		.colorSwatch {
+			position: absolute;
+			.position(0);
+			border-radius: @agate-colorpicker-swatch-border-radius;
+		}
+
+		.colorSwatch {
+			margin: @agate-colorpicker-swatch-spacing;
```
Djspaceg: Perhaps this should be more _direct_ and refer to `@agate-colorpicker-swatch-spacing` so it's a little clearer about where it's going and that it can support multiple values?

ColorPicker/SwatchButton.js
```diff
@@ -0,0 +1,134 @@
...
+	handlers: {
+		onClick: (ev, {children, color, onClick}) => {
+			if (onClick) {
+				onClick({color: color ? color : children});
+			}
+		}
+	},
+
+	computed: {
+		colorSwatchStyle: ({children, color}) => ({backgroundColor: color ? color : children})
```
Djspaceg: maybe more simply: `{backgroundColor: color || children}`.

We should probably also set a proptype for `children` to be a string.

Djspaceg:
For some reason, the drawer expands to whatever size it wants and also wraps the children so it looks a little funny at various widths. I tweaked the rounded corners a little to mask this fact, but we may still want to look into it later.

PR #77, opened 2018-11-17
PLAT-68551: Added a drawer component

Djspaceg:
It's basic, but it will be a fine starting point.
The `Icon` is included (but commented out), in case we want to eventually add a close/handle actuator.


Drawer/Drawer.js
```diff
@@ -0,0 +1,84 @@
...
+				duration="short"
+				type="slide"
+				className={css.drawerTransitionContainer}
+				onHide={onHide}
+				css={css}
+			>
+				<Layout
+					{...rest}
+				>
+					{/* <Icon small onClick={onClose} className={css.closeButton}>closex</Icon>*/}
```
germboy: What are your plans for this (cruft/future)? I assume this was/will-be a built-in feature to allow managing closed state. I'd generally suggest removing comments, but if we're wanting to keep this for a future feature, then perhaps we could also:
- comment out the `css.closeButton` styles in `Drawer.less` for now
- create/destructure and comment out the prop for `onClose`, since it's used without reference here (even though it's commented).

Djspaceg: Those are good points. I had indeed left this in for future consideration, something to the similar visual effect of the old Moonstone drawer, with the close button looking like a little tab in the center. Commenting out the other parts and including a forward-referencing comment/description explaining the intent is a great idea. I'll add that and make this aspect look a little more presentable. The locally controlled closing ability wasn't immediately needed for this scenario, but will surely be in the future.

PR #79, opened 2018-11-19
the smallest buttons!

sugardave:
This patch relies on https://github.com/enactjs/enact/pull/2024 being merged.

This adds the `smallest` rules for `Button` and `Icon` to the existing skins.


webOS101:
We took a different approach.

PR #83, opened 2018-12-12
PLAT-69282: Implement new UI theme

sugardave:
This adds the first designed theme, Copper.  It will serve as the basis for Copper Light and potentially other themes as time allows.


Djspaceg:
Used as the basis for our current copper implementation.

PR #89, opened 2018-12-21
PLAT-71492: Copper Day Theme

teckliew:
Copper Day theme.

To be merged with https://github.com/enactjs/agate-apps/pull/167
For sampler and console support of the theme

PLAT-71492


Button/Button.less
```diff
@@ -122,7 +122,7 @@
 
 			&,
 			.client {
-				color: @agate-button-selected-color;
```
Djspaceg: Was this intentional? Switching from "selected" to "highlighted" for the "selected" state? [Line 118](https://github.com/enactjs/agate/pull/89/files#diff-4caeea3683cfa034336602bc8f94573fL118)

TabGroup/TabGroup.js
```diff
@@ -35,21 +35,24 @@ const TabBase = kind({
 	},
 
 	computed: {
-		tabLabel: ({children, className, css, icon, labelPosition, orientation, selected, ...rest}) => {
-			let inline;
+		tabLabel: ({children, className, css, icon, labelPosition, orientation, index, ...rest}) => {
```
Djspaceg: I think I need a walkthrough of the intent of the changes to TabGroup before reviewing.

teckliew: Because my assumption was that `selected` is a boolean prop, I ran into a bit of confusion when `selected` was being used as if it's an `index` prop.
I went ahead with these changes to keep it consistent where `index` is the current `index` selected, and reserve `selected` as a boolean prop, which we do not need at the moment or AFAIK.

TabGroup/TabGroup.js
```diff
@@ -35,21 +35,24 @@ const TabBase = kind({
...
+
 			if (className.includes('copper')) {
 				return (
-					<div className={css.labeledIcon}>
-						<IconButton
+					<div className={selected ? `${css.selectedLabeledIcon} ${css.labeledIcon}` : css.labeledIcon}>
+						<ToggleButton
+							icon={icon}
 							inline={inline}
-							selected={selected === rest['data-index']}
```
teckliew: Here is where we used mix types for `selected`

TabGroup/TabGroup.js
```diff
@@ -35,21 +35,24 @@ const TabBase = kind({
...
-						>
-							{icon}
-						</IconButton>
+							selected={selected}
```
teckliew: Made changes so we can just do this.

TabGroup/TabGroup.js
Djspaceg: I see. maintaining consistency between components for prop names (Button, ToggleButton, etc use `selected` as a boolean). How would you feel about updating TabGroup to use `selectedIndex` rather than `index` so it's a little more descriptive?

I think that `Group` and `Repeater` have some props that let consumers more easily customize the prop name which may help reduce the number of changed lines here. Worth looking into. If the theme doesn't directly rely on these changes, we could put them in a separate PR to help compartmentalize the review. WDYT?

teckliew: Sounds good `selectedIndex` is better.

Panels/TabbedPanels.js
```diff
@@ -61,7 +62,8 @@ const TabbedPanelsBase = kind({
 						orientation={tabOrientation}
 						tabs={tabs}
 						tabPosition={tabPosition}
-						selected={index}
+						selectedIndex={selectedIndex}
+						index={index}
```
Djspaceg: Should this have both `index` and `selectedIndex`? We could probably simplify this to only need one of those two. I think `index` isn't handled anymore by TabGroup.

PR #93, opened 2018-12-27
PLAT-71634: Cleanup SwitchItem and Switch

Djspaceg:
Converted custom layout to Layout component for better reliability, ignore taps on the Switch when inside SwitchItem, convert onTaps to onClicks to avoid quirk in Tappable, general cleanup.

### DIFF
--- a/SwitchItem/SwitchItem.js
+++ b/SwitchItem/SwitchItem.js
@@ -7,18 +7,20 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
-
-import kind from '@enact/core/kind';
-import AgateSwitch from '../Switch';
-import Icon from '../Icon';
-import Skinnable from '../Skinnable';
-
-import Spottable from '@enact/spotlight/Spottable';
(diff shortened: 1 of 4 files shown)


SwitchItem/SwitchItem.js
```diff
@@ -100,32 +100,29 @@ const SwitchItemBase = kind({
...
 		);
 	}
 });
 
-
 const SwitchItemDecorator = compose(
-	Toggleable({toggleProp: 'onTap'}),
-	Touchable,
+	Toggleable({toggleProp: 'onClick'}),
+	// Touchable,
```
viodragon2: Do we need to remove `Touchable`?

Djspaceg: I did because it didn't seem to _help_, but it also didn't really seem to _hurt_ either? I don't feel strongly about its presence.

Djspaceg:
@viodragon2 :mergeitbuddy:!

PR #102, opened 2019-01-04
PLAT-72136: Added out-of-component click handler to dismiss the color picker palette

Djspaceg:


ColorPicker/ColorPicker.js
```diff
@@ -215,16 +219,41 @@ const ColorPickerExtended = hoc((config, Wrapped) => {
...
+
+			if (!prevProps.open && open) {
+				on('click', this.handleClick);
+			} else if (prevProps.open && !open) {
+				off('click', this.handleClick);
+			}
+		}
+
+		componentWillUnmount () {
+			off('click', this.handleClick);
```
germboy: I believe we only need to call this if we unmount while the popup is open, otherwise there's no handler attached

webOS101: There's no need to gate it, you can remove a handler, even if it doesn't exist, without error.

ColorPicker/ColorPicker.js
```diff
@@ -215,16 +219,41 @@ const ColorPickerExtended = hoc((config, Wrapped) => {
 			};
 		}
 
+		componentDidMount () {
+			this.node = ReactDOM.findDOMNode(this);
+		}
```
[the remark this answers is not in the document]
Djspaceg: Updated! Thanks. Nice find.

PR #109, opened 2019-06-18
PLAT-80254: Update Agate components to expose public class names

teckliew:
Update `publicClassNames` for `LabeledIcon`, `LabeledIconButton`, `Picker`, `ProgressBar`, `SliderButton`, and `ToggleButton`.

### DIFF
--- a/LabeledIcon/LabeledIcon.js
+++ b/LabeledIcon/LabeledIcon.js
@@ -29,9 +29,14 @@ const LabeledIconBase = kind({
 		children: PropTypes.oneOfType([PropTypes.string, PropTypes.object]).isRequired
 	},
 
+	styles: {
+		css: componentCss,
+		className: 'labeledIcon',
+		publicClassNames: true
+	},
+
 	render: (props) => UiLabeledIcon.inline({
 		...props,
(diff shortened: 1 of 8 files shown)


LabeledIcon/LabeledIcon.js
```diff
@@ -29,9 +29,14 @@ const LabeledIconBase = kind({
...
 
+	styles: {
+		css: componentCss,
+		className: 'labeledIcon',
+		publicClassNames: true
+	},
+
 	render: (props) => UiLabeledIcon.inline({
 		...props,
-		css: componentCss,
```
Djspaceg: To send the `css` prop down to `UiLabeledIcon` we must (always) manually extract it from the props and add it back. It's not included in the spread operation:
```
render: ({css, ...rest}) => UiLabeledIcon.inline({
	...rest,
	css,
```

ryanjduffy: Alternatively, you can skip deconstructing and reference it directly.
```
render: (props) => UiLabeledIcon.inline({
  ...props,
  css: props.css
});
```

PR #110, opened 2019-08-12
PLAT-84218: Update Agate to Enact 3.0

sugardave:
Update dependencies to Enact 3.0 and upgrade Agate components.


JayCanuck:
Re-targetted for new `develop` branch

PR #112, opened 2019-08-16
PLAT-84314: Add Searching Spinner

teckliew:
Add `Spinner` with `color` ("dark" and "light") and `size` ("small" and "medium") props.

spinner
Light Spinner

spinnerdark
Dark Spinner

### DIFF
--- a/Spinner/Spinner.js
+++ b/Spinner/Spinner.js
@@ -22,8 +22,6 @@ import Skinnable from '../Skinnable';
 
 import componentCss from './Spinner.module.less';
 
-const createSpinnerNodes = (numberOfNodes, css) => [...Array(numberOfNodes)].map((_, index) => <span className={css.fan + ' ' + css[`fan${index + 1}`]} key={`fan${index + 1}`} />);
-
 /**
  * A component that shows spinning fan.
  *
@@ -35,11 +33,19 @@ const createSpinnerNodes = (numberOfNodes, css) => [...Array(numberOfNodes)].map
 const SpinnerCore = kind({
 	name: 'SpinnerCore',
 
-	propTypes: {
+	propTypes: /** @lends agate/Spinner.SpinnerCore.prototype */ {
+		/**
+		 * The "aria-label" for the component.
+		 *
+		 * @type {String}
+		 * @public
+		 */
+		'aria-label': PropTypes.string,
+
 		css: PropTypes.object,
 
 		/**
-		 * Type of spinner.
+		 * The type of spinner.
 		 *
 		 * @type {('loading'|'searching')}
 		 * @default 'searching'
@@ -57,17 +63,29 @@ const SpinnerCore = kind({
 	},
 
 	computed: {
+		'aria-label': ({'aria-label': label, type}) => {
+			// TODO: These static values will need to be localized
+			return label || type === 'searching' ? 'Searching' : 'Loading';
+		},
 		className: ({styler, type}) => styler.append(type),
-		spinnerNodes: ({css, type}) => type === 'searching' ? createSpinnerNodes(12, css) : createSpinnerNodes(4, css)
+		spinnerNodes: ({styler, type}) => {
+			return Array.from({length: type === 'searching' ? 12 : 4}, (_, index) => (
+				<span className={styler.join('fan', `fan${index + 1}`)} key={`fan${index}`} />
+			));
+		}
 	},
 
-	render: ({css, spinnerNodes, type, ...rest}) => (
-		<div aria-label={type} aria-live="off" role="alert" {...rest}>
-			<div className={css.bg}>
-				{spinnerNodes}
+	render: ({css, spinnerNodes, ...rest}) => {
+		delete rest.type;
+
+		return (
+			<div aria-live="off" role="alert" {...rest}>
+				<div className={css.bg}>
+					{spinnerNodes}
+				</div>
 			</div>
-		</div>
-	)
+		);
+	}
 });
 
 /**
@@ -83,7 +101,7 @@ const SpinnerBase = kind({
 
 	propTypes: /** @lends agate/Spinner.SpinnerBase.prototype */ {
 		/**
-		 * Customize the color of spinner.
+		 * The color of the component.
 		 *
 		 * @type {('dark'|'light')}
 		 * @default 'light'
@@ -97,7 +115,8 @@ const SpinnerBase = kind({
 		 *
 		 * The following classes are supported:
 		 *
-		 * * `spinner` - The root component class, unless there is a scrim. The scrim and floating layer can be a sibbling or parent to this root "spinner" element.
+		 * * `spinner` - The root component class, unless there is a scrim. The scrim and floating
+		 *   layer can be a sibbling or parent to this root "spinner" element.
 		 *
 		 * @type {Object}
 		 * @public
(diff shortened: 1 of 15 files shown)


styles/colors-carbon.less
```diff
@@ -244,6 +244,12 @@
 @agate-sliderbutton-knob-shadow:    @agate-slider-knob-shadow;
 @agate-sliderbutton-knob-focus-shadow: @agate-slider-knob-focus-shadow;
 
+// Spinner
+// ---------------------------------------
+@agate-spinner-light-color: @agate-white;
+@agate-spinner-dark-color: @agate-darker-gray;
+@agate-spinner-scrim-color: fade;
```
Djspaceg: What does `background-color: fade;` do?

Spinner/Spinner.module.less
```diff
@@ -0,0 +1,196 @@
...
+			.fan7,
+			.fan8,
+			.fan9,
+			.fan10,
+			.fan11,
+			.fan12 {
+				transform-origin: @agate-spinner-small-transform-origin @agate-spinner-small-transform-origin;
+
+				&:after {
+					left: 16px;
```
Djspaceg: Will these values always be the same for all spinners in all themes?

Spinner/Spinner.module.less
```diff
@@ -0,0 +1,196 @@
+// Spinner.module.less
+//
+@import "../styles/mixins.less";
+@import "../styles/variables.less";
+@import "../styles/skin.less";
+
+@keyframes spinner {
```
Djspaceg: We'll likely want a more specific name here, since there may be multiple types of spinners.

Spinner/Spinner.module.less
```diff
@@ -0,0 +1,196 @@
...
+}
+
+.spinner {
+	.bg {
+		display: inline-block;
+		width: @agate-spinner-size;
+		height: @agate-spinner-size;
+		position: relative;
+
+		.fan1,
```
Djspaceg: We can simplify several of these repetitive sections if the "blades" have two classes, or we use the next sibling selector `+`. E.g.: 
```
.blade {}
.blade + .blade {}
.blade + .blade + .blade {}
// Alternatively
.blade:nth-child(1) {}
.blade:nth-child(2) {}
.blade:nth-child(3) {}
```

Spinner/Spinner.module.less
```diff
@@ -0,0 +1,196 @@
...
+			transform-origin: @agate-spinner-transform-origin @agate-spinner-transform-origin;
+			will-change: opacity;
+			animation: spinner 1.2s linear infinite;
+
+			&:after {
+				content: " ";
+				display: block;
+				position: absolute;
+				top: 0;
+				left: 23px;
```
Djspaceg: Let's have these values (this one and the two following) be relative or calculated based on the size of the spinner for some more versatility.

Spinner/Spinner.module.less
```diff
@@ -0,0 +1,196 @@
...
+		.fan9,
+		.fan10,
+		.fan11,
+		.fan12 {
+			transform-origin: @agate-spinner-transform-origin @agate-spinner-transform-origin;
+			will-change: opacity;
+			animation: spinner 1.2s linear infinite;
+
+			&:after {
+				content: " ";
```
Djspaceg: Is a space character necessary or could this be simply `""`. I ask because by inserting a space glyph, we're actually imposing text sizing boxes into this, when that may not be intended.

Spinner/Spinner.js
```diff
@@ -0,0 +1,183 @@
...
+	},
+
+	styles: {
+		css: componentCss
+	},
+
+	render: ({css, ...rest}) => (
+		<div aria-label="Loading" aria-live="off" role="alert" {...rest}>
+			<div className={css.bg}>
+				<div className={css.fan1} />
```
Djspaceg: It may be a good idea to have these nodes be generated via a `map`. That way, a computed property could decide the about of elements to create and we could easily support spinners with different amounts of nodes inside. Something like: 
```
computed: {
    spinnerNodes: ({type}) => (type === 'searching' ? 4 : 12)
}
```

Spinner/Spinner.module.less
```diff
@@ -0,0 +1,196 @@
...
+@import "../styles/mixins.less";
+@import "../styles/variables.less";
+@import "../styles/skin.less";
+
+@keyframes spinner {
+	0% {
+		opacity: 1;
+	}
+	100% {
+		opacity: 0;
```
Djspaceg: Does it actually fade all the way out to 0? It didn't look like it to me, but you looked at it a lot closer.

Spinner/Spinner.js
```diff
@@ -57,17 +63,29 @@ const SpinnerCore = kind({
...
 	},
 
-	render: ({css, spinnerNodes, type, ...rest}) => (
-		<div aria-label={type} aria-live="off" role="alert" {...rest}>
-			<div className={css.bg}>
-				{spinnerNodes}
+	render: ({css, spinnerNodes, ...rest}) => {
+		delete rest.type;
+
+		return (
```
ryanjduffy: Although it is more terse to "return" the JSX from the arrow fn, we usually have to change it later to remove props so it's often better to do so initially so the later diffs are more clear.

Spinner/Spinner.js
```diff
@@ -57,17 +63,29 @@ const SpinnerCore = kind({
...
 
 	computed: {
+		'aria-label': ({'aria-label': label, type}) => {
+			// TODO: These static values will need to be localized
+			return label || type === 'searching' ? 'Searching' : 'Loading';
+		},
 		className: ({styler, type}) => styler.append(type),
-		spinnerNodes: ({css, type}) => type === 'searching' ? createSpinnerNodes(12, css) : createSpinnerNodes(4, css)
+		spinnerNodes: ({styler, type}) => {
+			return Array.from({length: type === 'searching' ? 12 : 4}, (_, index) => (
```
ryanjduffy: Couple changes here:

* Since the spinner nodes are only needed for this computed, including the code inline here (vs refactored into a separate method) improves the readability of the component
* Using `Array.from()` can simplify creating an array of components since it accepts a map fn as its second arg

Spinner/Spinner.js
```diff
@@ -57,17 +63,29 @@ const SpinnerCore = kind({
 	},
 
 	computed: {
+		'aria-label': ({'aria-label': label, type}) => {
+			// TODO: These static values will need to be localized
+			return label || type === 'searching' ? 'Searching' : 'Loading';
```
ryanjduffy: We'll eventually need to support localization so I've refactored this into a computed prop rather than assigning the `type` directly as `aria-label`.

Spinner/Spinner.js
```diff
@@ -127,12 +146,14 @@ const SpinnerBase = kind({
...
 	defaultProps: {
+		color: 'light',
 		size: 'medium',
 		transparent: false
 	},
 
 	styles: {
-		css: componentCss
+		css: componentCss,
+		publicClassNames: ['spinner']
```
ryanjduffy: The `css` prop only documents `'spinner'` as a supported override so I've added it here. Without this, we would be supporting all the overrides for `ui/Spinner` since we pass `css` down to it.

Spinner/Spinner.js
```diff
@@ -143,17 +164,15 @@ const SpinnerBase = kind({
...
+	render: ({css, ...rest}) => {
 		delete rest.transparent;
 
 		return (
 			<UiSpinnerBase
 				{...rest}
 				css={css}
 				component={SpinnerCore}
-			>
-				{children}
```
ryanjduffy: I removed `children` since it is not currently supported by agate and is automatically handled by including it in `rest`.

PR #115, opened 2019-08-19
PLAT-84119: Improve Panels Support

germboy:
Adding support more features in `Panels`, including:
- controls (app close button)
- breadcrumbs (included in a specific `ActivityPanels` component)


Panels/IdProvider.js
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * A higher-order component that generates globally-unique identifiers
+ *
+ * @class IdProvider
+ * @hoc
+ * @private
+ * @memberof agate/Panels
+ */
+const IdProvider = hoc(defaultConfig, (config, Wrapped) => {
```
sugardave: We talked in Slack about this and I agree this could be moved somewhere else and refactor agate and moonstone to use those, like `ui` or maybe `core`?

Panels/SharedStateDecorator.js
```diff
@@ -0,0 +1,150 @@
...
+ * initiate an update cycle in React. The intent is only to restore state on mount.
+ *
+ * If shared state is used in the render method for a component, it may be necessary to use the
+ * `updateOnMount` config member which will initiate an update cycle within React once the data is
+ * available from an upstream shared state.
+ *
+ * @hoc
+ * @private
+ */
+const SharedStateDecorator = hoc(defaultConfig, (config, Wrapped) => {
```
sugardave: Like `IdProvider` but we might have a deeper team discussion about whether agate should provide this functionality.

Panels/Panels.js
```diff
@@ -38,25 +157,78 @@ const PanelsBase = kind({
...
+const PanelsDecorator = compose(
+	Slottable({slots: ['controls']}),
+	CancelDecorator({cancel: 'onBack'}),
+	Measurable({refProp: 'controlsRef', measurementProp: 'controlsMeasurements'}),
+	IdProvider
+);
+
+const Panels = PanelsDecorator(PanelsBase);
+
 export default PanelsBase;
```
sugardave: Should the default export be `Panels` now and only let devs get `PanelsBase` by name?

Panels/Panels.js
```diff
@@ -38,25 +157,78 @@ const PanelsBase = kind({
...
 export {
 	Panel,
```
sugardave: Should we move this import/export to `index.js` with the other ones?

Panels/Panels.js
germboy: Good observation 👍

Panels/Panels.js
germboy: agreed - I missed this in my initial pass-through 🙈

Panels/IdProvider.js
germboy: Yeah, we'll need to ultimately do some module organization. It looks like for the time being, we'll bundle them in agate until we are able to introduce enact API changes. I've left these alone for now.

Panels/Viewport.js
```diff
@@ -0,0 +1,235 @@
...
+		 * When using `'vertical'`, a `SlideBottomArranger` is used.
+		 *
+		 * Valid values are:
+		 * * `'horizontal'`, and
+		 * * `'vertical'`.
+		 *
+		 * @type {String}
+		 * @public
+		 */
+		orientation: PropTypes.string
```
ryanjduffy: WDYT of removing this prop and only using the `arranger` prop. It doesn't add a lot of value as only a proxy for selecting the arrangers?

Djspaceg: History: It was added specifically for that reason, to be a simpler way to adjust Panels with just a single string prop change, rather than importing multiple arrangers at app level and swapping them out during runtime. With the CES demo, the app needed to be able to conveniently switch between orientations, so it made sense at the time. It's probably not the best approach.

PR #116, opened 2019-08-19
PLAT-84544: Add Header

teckliew:
Add basic Header with subtitle below and title above.
Sample: https://github.com/enactjs/agate-apps/pull/215
agate-header

### DIFF
--- a/Header/Header.js
+++ b/Header/Header.js
@@ -14,7 +14,7 @@
 import kind from '@enact/core/kind';
 import React from 'react';
 import PropTypes from 'prop-types';
-import {Layout, Cell} from '@enact/ui/Layout';
+import {Column, Row, Layout} from '@enact/ui/Layout';
 import Slottable from '@enact/ui/Slottable';
 
 import Skinnable from '../Skinnable';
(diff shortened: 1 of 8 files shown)


Header/Header.js
```diff
@@ -0,0 +1,143 @@
...
+		delete rest.titleAbove;
+
+		return (
+			<Row component="header" aria-label={title} {...rest}>
+				<Column className={css.titleContainer}>
+					{titleAboveComponent}
+					<h1 className={css.title}>{title}</h1>
+					{subtitleComponent}
+				</Column>
+				{children ? <Layout className={css.endSlot} shrink>{children}</Layout> : null}
```
germboy: The usage of `shrink` here is causing a prop bleed error when attempting to render `children`, since `shrink` is a property of `Cell`, not `Layout`. We may be able to just remove the prop altogether, unless there was a reason for the initial use of it. I think things would be ok without it.

teckliew: Yea that was cruft when it was a `Cell`. I can remove the prop.

PR #120, opened 2019-08-23
PLAT-84903: Add Galluim theme

Djspaceg:
Also updated applySkins to allow the assignment of a "base" skin, which other skins can leverage for overrides, so they don't all need to be updated when a single variable is added.


ryanjduffy:
Are the two Button changes cruft?

Djspaceg:
"Cruft" may be harsh. Not _directly_ related to this task, sure, I'll grant that. One was, I think, due to trying to improve compatibility with `ui`, since I think `ui` was exporting a non-kind-generated component for Button, so it didn't have inline any more. The other, backdrop-filter, was to implement the feature _somewhere_ and behind a transparent button seemed to make the most sense. I'd included them in the first commit of this when I was expecting this to be more of a "cleanup general things" PR.

ryanjduffy:
`ui/Button.ButtonBase` is a `kind` component so `inline()` should be safe. Okay with reverting that part? Not worried about the backdrop filter change.

Djspaceg:
Yes, we can drop all of the changes to that file. It adds little-to-no value.

PR #125, opened 2019-09-04
PLAT-84726: Add Heading component

Djspaceg:
This adds a new Heading component, with primary working being added for the "title" size of the component.

### DIFF
--- a/Heading/Heading.js
+++ b/Heading/Heading.js
@@ -1,0 +1,163 @@
+/**
+ * Agate styled labeled Heading components and behaviors
+ *
+ * @example
+ * <Heading
+ *   size="large"
+ *   spacing="small"
+ * >
+ *   A Content Section Heading
+ * </Heading>
+ *
+ * @module agate/Heading
+ * @exports Heading
+ * @exports HeadingBase
+ * @exports HeadingDecorator
+ */
+
+import kind from '@enact/core/kind';
+import Pure from '@enact/ui/internal/Pure';
+import PropTypes from 'prop-types';
+import compose from 'ramda/src/compose';
+import UiHeading from '@enact/ui/Heading';
+
+import Skinnable from '../Skinnable';
+
+import componentCss from './Heading.module.less';
+
+/**
+ * A labeled Heading component.
+ *
+ * This component is most often not used directly but may be composed within another component as it
+ * is within [Heading]{@link agate/Heading.Heading}.
+ *
+ * @class HeadingBase
+ * @memberof agate/Heading
+ * @ui
+ * @public
+ */
+const HeadingBase = kind({
+	name: 'Heading',
+
+	propTypes: /** @lends agate/Heading.HeadingBase.prototype */ {
+		css: PropTypes.object,
+
+		/**
+		 * Set a custom color to be used by the Heading. Defaults to the current accent color if
+		 * none is specified.
+		 *
+		 * @type {String}
+		 * @default null
+		 * @public
+		 */
+		customColor: PropTypes.string,
+
+		/**
+		 * Adds a horizontal-rule (line) under the component
+		 *
+		 * @type {Boolean}
+		 * @public
+		 */
+		showLine: PropTypes.bool,
+
+		/**
+		 * The size of the spacing around the Heading.
+		 *
+		 * Allowed values include:
+		 * * `'auto'` - Value is based on the `size` prop for automatic usage.
+		 * * `'large'` - Specifically assign the `'large'` spacing.
+		 * * `'medium'` - Specifically assign the `'medium'` spacing.
+		 * * `'small'` - Specifically assign the `'small'` spacing.
+		 * * `'none'` - No spacing at all. Neighboring elements will directly touch the Heading.
+		 *
+		 * @type {('auto'|'title'|'large'|'medium'|'small'|'none')}
+		 * @default 'small'
+		 * @public
+		 */
+		spacing: PropTypes.oneOf(['auto', 'title', 'large', 'medium', 'small', 'none'])
+	},
+
+	defaultProps: {
+		spacing: 'small'
+	},
+
+	styles: {
+		css: componentCss,
+		className: 'heading'
+	},
+
+	computed: {
+		className: ({showLine, styler}) => styler.append({showLine}),
+		style: ({customColor, style}) => ({
+			...style,
+			'--agate-heading-accent': customColor
+		})
+	},
+
+	render: ({css, ...rest}) => {
+		delete rest.customColor;
+		delete rest.showLine;
+		return UiHeading.inline({css, ...rest});
+	}
+});
+
+/**
+ * Applies Agate specific behaviors to [HeadingBase]{@link agate/Heading.HeadingBase}.
+ *
+ * @hoc
+ * @memberof agate/Heading
+ * @mixes agate/Marquee.MarqueeDecorator
+ * @mixes agate/Skinnable.Skinnable
+ * @public
+ */
+const HeadingDecorator = compose(
+	Pure,
+	Skinnable
+);
+
+/**
+ * A labeled Heading component, ready to use in Agate applications.
+ *
+ * `Heading` may be used as a header to group related components.
+ *
+ * Usage:
+ * ```
+ * <Heading
+ *   spacing="medium"
+ * >
+ *   Related Settings
+ * </Heading>
+ * <CheckboxItem>A Setting</CheckboxItem>
+ * <CheckboxItem>A Second Setting</CheckboxItem>
+ * ```
+ *
+ * @class Heading
+ * @memberof agate/Heading
+ * @extends agate/Heading.HeadingBase
+ * @mixes agate/Heading.HeadingDecorator
+ * @ui
+ * @public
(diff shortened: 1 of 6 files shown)


ryanjduffy:
Should this supersede the divider component that's in agate now?

Djspaceg:
Yep. It sure should. Let's make a ticket for that deprecation and we can follow up soon.

Heading/Heading.js
```diff
@@ -0,0 +1,163 @@
...
+
+		/**
+		 * Set a custom color to be used by the Heading. Defaults to the current accent color if
+		 * none is specified.
+		 *
+		 * @type {String}
+		 * @default null
+		 * @public
+		 */
+		customColor: PropTypes.string,
```
ryanjduffy: Why did you include the "custom" prefix here? That word doesn't carry a lot of meaning for me in this context since any color I'd pass would be "custom". Maybe just `color` or `accentColor` if the purpose of the color needs to be included in the name. Thoughts?

Heading/Heading.js
```diff
@@ -0,0 +1,163 @@
...
+		return UiHeading.inline({css, ...rest});
+	}
+});
+
+/**
+ * Applies Agate specific behaviors to [HeadingBase]{@link agate/Heading.HeadingBase}.
+ *
+ * @hoc
+ * @memberof agate/Heading
+ * @mixes agate/Marquee.MarqueeDecorator
```
ryanjduffy: Let's remove this for now.

Heading/Heading.js
```diff
@@ -0,0 +1,163 @@
...
+ * @class Heading
+ * @memberof agate/Heading
+ * @extends agate/Heading.HeadingBase
+ * @mixes agate/Heading.HeadingDecorator
+ * @ui
+ * @public
+ */
+const Heading = HeadingDecorator(HeadingBase);
+
+/**
```
ryanjduffy: Let's remove this block as well.

Heading/Heading.js
Djspaceg: I'd originally started with just "color", but I didn't feel that conveyed enough contextual meaning. Implying that without some qualifier, there would be no color. There is in fact a default color, which itself is the accent color. I didn't choose the logical "accentColor" because I didn't want confusion about whether this was overriding/updating the skin's accent color or just applying to this component. "customColor" was the compromise which indicates that you're customizing the color beyond its existing, default color. I can change it back to `color` though, I _think_ I'd be ok with that.

Heading/Heading.js
Djspaceg: oof forgot that little guy.

Heading/Heading.module.less
```diff
@@ -0,0 +1,86 @@
...
+			position: relative;
+			padding: @agate-heading-title-padding;
+
+			&::after {
+				content: "";
+				position: absolute;
+				height: @agate-heading-title-decoration-height;
+				width: @agate-heading-title-decoration-width;
+				.position(@agate-heading-title-decoration-positions);
+				background-color: var(--agate-heading-accent, @agate-heading-accent-color);
```
ryanjduffy: `@agate-heading-accent-color` isn't defined in the colors file

Djspaceg: Crap... it's defined in the other ticket. :ughwhy:. I'll import it.

Heading/Heading.js
```diff
@@ -0,0 +1,147 @@
...
+		 *
+		 * @type {('auto'|'title'|'large'|'medium'|'small'|'none')}
+		 * @default 'small'
+		 * @public
+		 */
+		spacing: PropTypes.oneOf(['auto', 'title', 'large', 'medium', 'small', 'none'])
+	},
+
+	defaultProps: {
+		spacing: 'small'
```
sugardave: Why not have the default be `auto`?

styles/variables-gallium.less
```diff
@@ -90,6 +90,41 @@
...
 
+// Heading
+// ---------------------------------------
+@agate-heading-border-width: 3px;
+@agate-heading-padding: 24px;
+@agate-heading-showline-padding: 12px;
+@agate-heading-letter-spacing: normal;
+@agate-heading-line-height: 1.6em;
+@agate-heading-title-font-family: @agate-base-font-family;
+@agate-heading-title-font-size: 70px;
```
sugardave: Should this font size be larger than `@agate-heading-large-font-size` (48px)?

styles/variables-gallium.less
```diff
@@ -90,6 +90,41 @@
...
+@agate-heading-medium-font-size: 32px;
+@agate-heading-medium-font-style: normal;
+@agate-heading-medium-font-weight: bold;
+@agate-heading-small-font-family: @agate-base-font-family;
+@agate-heading-small-font-size: 24px;
+@agate-heading-small-font-style: normal;
+@agate-heading-small-font-weight: bold;
+@agate-heading-spacing-title: 32px;
+@agate-heading-spacing-large: 24px;
+@agate-heading-spacing-medium: 24px;
```
sugardave: Same as `@agate-heading-spacing-large`?

Heading/Heading.module.less
```diff
@@ -0,0 +1,86 @@
...
+
+		&.title {
+			font-family: @agate-heading-title-font-family;
+			font-weight: @agate-heading-title-font-weight;
+			font-style: @agate-heading-title-font-style;
+			font-size: @agate-heading-title-font-size;
+			position: relative;
+			padding: @agate-heading-title-padding;
+
+			&::after {
```
ryanjduffy: Suggest we guard this with `showLine` as well for consistency across the sizes.

```less
&.showLine {
  border-bottom-color: transparent;

  &::after {
    /* rest of the rules */
  }
}
```

Heading/Heading.module.less
```diff
@@ -0,0 +1,86 @@
...
+		border-bottom: @agate-heading-border-width solid transparent;
+		margin: 0 @agate-spotlight-outset (1.5 * @agate-spotlight-outset) @agate-spotlight-outset;
+		padding: @agate-heading-padding;
+		min-height: (1.5 * @agate-spotlight-outset);
+		letter-spacing: @agate-heading-letter-spacing;
+		line-height: @agate-heading-line-height;
+		color: @agate-heading-text-color;
+
+		&.showLine {
+			border-bottom-color: @agate-heading-border-color;
```
ryanjduffy: Suggest we respect the `customColor`/`color` prop here as well.

ryanjduffy: If we choose to do this, the `@agate-heading-accent-color` may not be necessary since it would fully overlap in purpose with the border color.

Heading/Heading.js
ryanjduffy: I'd be in favor of `color` but `lineColor` might be another option if we support it for the "border" color across sizes.

Heading/Heading.js
Djspaceg: That's a good question. It probably should be.

styles/variables-gallium.less
Djspaceg: Indeed it should! Title -> SubTitle -> Large -> Medium -> Small -> Tiny, which relate to `h1` -> `h6`. Clever eh? :D

styles/variables-gallium.less
Djspaceg: Ditto :D

Heading/Heading.module.less
Djspaceg: I'd considered that too. I was thinking that the "line" was separate from the "decoraton", and you may want both. I didn't see any designs in the doc that had a "page title" level header that did not include a decoration (the little colored block), but also figured that since the "showLine" implied a known entity, like a line that went all the way across the space, like an HR, that they should be interpreted separately. I can conceede this point though as both arguments are logical. Having the component behave more intuitively to a new user (rather than the author) is higher priority.

Heading/Heading.js
Djspaceg: I was wanting to avoid visual terminology so a skin could determine the shape for itself. It's a line in gallium, but it's also slanted. Other skins may leverage this feature differently.

Heading/Heading.module.less
Djspaceg: The `showLine` prop is false by default, being a boolean. The decoration part should be showing by default for title-sized Headings. I'm really inclined to leave these two separate for the time being.

Heading/Heading.module.less
Djspaceg: I initially agreed with this, but after implementing it, I am having second thoughts. I'm thinking of a scenario where the custom color is set, and the showLine and the decorator are both showing (the showLine line demarcating the content below the title and the decorator adding visual styling to the title text), there would be no way to have the two be different colors since they're both based on the prop.

Heading/Heading.js
Djspaceg: I updated it to `color` for simplicity. Moonstone Button uses plain `color` prop, so this could too I suppose.

Djspaceg:
PR has been updated with almost all of the suggestions, but I'm content with moving it along now unless there are any new issues.

PR #126, opened 2019-09-04
PLAT-85175: Updated Icon and IconButton with huge size, Button with new colors

Djspaceg:
This includes the "day" version of gallium and layers onto the Heading ticket #125, including colors for Heading in gallium-day.

`Button` received a new prop for setting a specific background color
`Icon` and `IconButton` now support a huge size, and leverage existing agate work rather than creating their own versions of things
Gallium-day has been added and is now the default style. (The settings app should be set explicitly to `skin="gallium"` so it gets the dark version of gallium.

### DIFF
--- a/Button/Button.js
+++ b/Button/Button.js
@@ -43,6 +43,8 @@ const ButtonBase = kind({
 	name: 'Button',
 
 	propTypes: /** @lends agate/Button.ButtonBase.prototype */ {
+		backgroundOpacity: PropTypes.oneOf(['opaque', 'lightOpaque']),
+
 		/**
 		 * Customizes the component by mapping the supplied collection of CSS class names to the
 		 * corresponding internal Elements and states of this component.
@@ -108,6 +110,7 @@ const ButtonBase = kind({
 	},
 
 	defaultProps: {
+		backgroundOpacity: 'opaque',
 		type: 'standard',
 		size: 'large'
 	},
@@ -118,7 +121,8 @@ const ButtonBase = kind({
 	},
 
 	computed: {
-		className: ({highlighted, joinedPosition, selected, type, styler}) => styler.append(
+		className: ({backgroundOpacity, highlighted, joinedPosition, selected, type, styler}) => styler.append(
+			backgroundOpacity,
 			type,
 			(joinedPosition && 'joined' + cap(joinedPosition)),  // If `joinedPosition` is present, prepend the word "joined" to the variable, so the classes are clearer.
 			{highlighted, selected}
@@ -127,6 +131,7 @@ const ButtonBase = kind({
 	},
 
 	render: ({css, ...rest}) => {
+		delete rest.backgroundOpacity;
 		delete rest.highlighted;
 		delete rest.joinedPosition;
 		delete rest.selected;
--- a/Heading/Heading.js
+++ b/Heading/Heading.js
@@ -1,0 +1,163 @@
+/**
+ * Agate styled labeled Heading components and behaviors
+ *
+ * @example
+ * <Heading
+ *   size="large"
+ *   spacing="small"
+ * >
+ *   A Content Section Heading
+ * </Heading>
+ *
+ * @module agate/Heading
+ * @exports Heading
+ * @exports HeadingBase
+ * @exports HeadingDecorator
+ */
+
+import kind from '@enact/core/kind';
+import Pure from '@enact/ui/internal/Pure';
+import PropTypes from 'prop-types';
+import compose from 'ramda/src/compose';
+import UiHeading from '@enact/ui/Heading';
+
+import Skinnable from '../Skinnable';
+
+import componentCss from './Heading.module.less';
+
+/**
+ * A labeled Heading component.
+ *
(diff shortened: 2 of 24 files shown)


Heading/Heading.js
```diff
@@ -0,0 +1,163 @@
+/**
```
sugardave: Just a reminder to make sure that changes in #125 get populated here or override these `Heading` files.

IconButton/IconButton.js
```diff
@@ -12,35 +12,18 @@
...
 import React from 'react';
 
-import AgateIcon from '../Icon';
+import {ButtonBase, ButtonDecorator} from '../Button';
+import Icon from '../Icon';
 import Skinnable from '../Skinnable';
 
 import componentCss from './IconButton.module.less';
 
 // Make a basic Icon in case we need it later. This cuts `Pure` out of icon for a small gain.
```
sugardave: Comment no longer applies, I think?

IconButton/IconButton.module.less
```diff
@@ -65,6 +65,12 @@
 			}
 		}
 
+		&.huge {
+			height: @agate-iconbutton-huge-height;
```
sugardave: I know this is "late to the party" for agate, but wondering why `IconButton` variables use "height" for sizing while `Icon` variables use "size"?  Comparing to moonstone, the same descriptor ("height") is used for both.

Djspaceg: Let's talk about Agate IconButton offline. :) Basically, it just hasn't seen the level of maintinence and tidying that the other components have.

Heading/Heading.js
Djspaceg: Merged in #125 after it was updated most recently.

Button/Button.js
```diff
@@ -43,6 +43,8 @@ const ButtonBase = kind({
 	name: 'Button',
 
 	propTypes: /** @lends agate/Button.ButtonBase.prototype */ {
+		backgroundOpacity: PropTypes.oneOf(['opaque', 'lightOpaque']),
```
ryanjduffy: Why did you choose these values as compared to `moonstone` which used an unset value as `opaque` and set values as degrees of opacity?

Button/Button.js
```diff
@@ -118,7 +121,8 @@ const ButtonBase = kind({
 	},
 
 	computed: {
-		className: ({highlighted, joinedPosition, selected, type, styler}) => styler.append(
+		className: ({backgroundOpacity, highlighted, joinedPosition, selected, type, styler}) => styler.append(
+			backgroundOpacity,
```
ryanjduffy: If we're going to pass this through to styler, we should define the class in the CSS even if it is unvalued. Otherwise, styler will treat it as a global class and apply it as is which could have unexpected side effects for authors.

styles/skin.less
```diff
@@ -6,7 +6,7 @@
 	.buildSkinRules(@skin; @importRules) {
 		&:global(.@{skin}) {
 			// Establish a base set of colors and variables that other skins can optionally override.
-			@import "./colors-gallium.less";
+			@import "./colors-gallium-day.less";
 			@import "./variables-gallium.less";
```
ryanjduffy: Should this be gallium-day as well?

Button/Button.js
Djspaceg: I'd say this is a case of gradually evolving preference for what my opinion is on "best" API. By using blank as the default, it means that, in the CSS, the rules must be overridden to set something else and you must override them, even if you just want to remove those rules. By setting up the API to have a default class, the CSS could be structured to work in both ways, depending on what suits the requirements, and not need to change the API later.

As for `lightOpaque`, I imagine there will likely be other levels of background opacity in the future, but this one specifically sets the background to white for some of the design's specific use cases. It could have been considered a "skin", however, it seemed like both day and night gallium designs used gray buttons as default, while occasionally needing white buttons in places. Perhaps Gallium needs a neutral or translucent variant for these cases, but that seemed premature at this point.

Button/Button.js
Djspaceg: Oh good point.

styles/skin.less
Djspaceg: `variables` is independent of day/night; there's just the one.

ryanjduffy: Understood. Might have a slight preference for making `gallium-day` be `gallium` if it's the default for consistent naming but not critical to resolve at the moment.

Djspaceg: Personally, I'd lean toward renaming `gallium` to `gallium-night` so it's clearer.

Button/Button.js
ryanjduffy: I think we need to be very careful of incrementally adopting API differences like this as it runs the risk of creating divergent API semantics over time.

Djspaceg: We use this approach in many parts of UI and Moonstone. Button is a legacy component at this point.

styles/skin.less
ryanjduffy: Whichever is fine

PR #127, opened 2019-09-04
PLAT-84897: Add PopupMenu component

Djspaceg:
A fresh new component for displaying simple quick lists over the rest of the content.

This PR depends on #125 and #126


PopupMenu/PopupMenu.js
```diff
@@ -0,0 +1,90 @@
...
+				noAnimation={noAnimation}
+				visible={open}
+				direction="down"
+				duration="short"
+				type="fade"
+				className={css.popupTransitionContainer}
+				onHide={onHide}
+				css={css}
+			>
+				<Layout orientation="vertical" align="center center" {...rest}>
```
sugardave: Are you using `Layout` instead of `Column` for possible future horizontal support?

PopupMenu/PopupMenuState.js
```diff
@@ -0,0 +1,92 @@
...
+		componentWillReceiveProps (nextProps) {
+			if (!this.props.open && nextProps.open) {
+				this.setState({
+					popupOpen: nextProps.noAnimation,
+					floatLayerOpen: true
+				});
+			} else if (this.props.open && !nextProps.open) {
+				this.setState({
+					popupOpen: nextProps.noAnimation,
+					floatLayerOpen: !nextProps.noAnimation
```
sugardave: Ouch, this makes my brain hurt!  Why isn't it just `false` for `floatLayerOpen`?

PopupMenu/PopupMenuState.js
```diff
@@ -0,0 +1,92 @@
...
+				});
+			}
+		}
+
+		handlePopupHide = () => {
+			forwardHide(null, this.props);
+
+			this.setState({
+				floatLayerOpen: false,
+				activator: null
```
sugardave: What is the purpose of `activator`?

PopupMenu/PopupMenu.js
Djspaceg: That was what I was thinking, but I didn't have a practical reason to arbitrarily switch the orientation in one skin or another, so I left it hardcoded at the moment with capability for override by `{...rest}`.

Fun fact: `Layout` is one less function call than `Column`, since `Column` is just a function that sets the `orientation`, so it's a tiny bit more efficient.

PopupMenu/PopupMenuState.js
Djspaceg: Oof! That whole file got unintentionally committed. It's not even referenced!

PR #128, opened 2019-09-05
ENYO-6245: agate/Heading: styles for spacing prop are not applied

sugardave:
Spacing styles now get added to `className`


sugardave:
I was mistaken about where spacing is applied, abandoning this :pr:

PR #133, opened 2019-09-17
ENYO-6275: Fixed `noCloseButton` prop to pass TabbedPanels to panels

Baekwoo-s:
There is no way to remove the close button of a TabbedPanels from app.
When this PR is merged, we can use `noCloseButton` prop to hide the close button on the TabbedPanels.


ybsung:
LGTM

PR #134, opened 2019-09-17
PLAT-85937: Freeze state in dev mode to prevent direct mutation

ryanjduffy:
We had some code that was directly mutating state to cache service requests. We've discussed other ways to handle that particular concern but this PR freezes the state object in dev-mode to prevent that mutation.


Djspaceg:
P.S. Is this the correct ticket number? I'm not seeing this PR's URL in that ticket (and it's closed).

ryanjduffy:
Nope :/ Fixed now!

PR #138, opened 2019-09-18
PLAT-84631: Consolidate item logic

ryanjduffy:


styles/variables-gallium.less
```diff
@@ -174,13 +174,11 @@
...
-@agate-item-font-weight: 300;
-@agate-item-label-font-size: @agate-base-font-size;
-@agate-item-height: 24px;
-@agate-item-side-outset: 12px;
-@agate-item-side-outset: 12px;
-@agate-item-height: 60px;
+@agate-item-font-size: 40px;
+@agate-item-font-weight: 500;
+@agate-item-label-font-size: 30px;
+@agate-item-side-outset: 32px;
```
ryanjduffy: This seems a bit large to me. Was this pulled from the specs?

Item/Item.module.less
```diff
@@ -16,13 +63,49 @@
...
+				font-size: @agate-item-label-font-size;
+			}
+		}
+
+		&.active {
+			color: @agate-item-active-color;
+		}
+
+		&.selected {
+			outline: @agate-item-selected-outline;
```
ryanjduffy: I noticed a faint gap in Chrome with device emulation when using `outline`. `box-shadow` doesn't seem to produce the same result so it might be a layout-preserving alternative.

image

Item/Item.module.less
```diff
@@ -4,10 +4,57 @@
...
+
+			&.labelAbove,
+			&.labelBefore {
+				.content { order: 2; }
+				.label   { order: 1; }
+			}
+
+			&.labelBefore .label {
+				// temp static values
+				.padding-start-end(0, 8px);
```
ryanjduffy: Let's move these into variables and remove my comment above.

Item/Item.module.less
```diff
@@ -4,10 +4,57 @@
...
+			&.labelAfter .label {
+				// temp static values
+				.padding-start-end(8px, 0);
+			}
+
+			.content {
+				box-sizing: border-box;
+				// when showing content and label inline, favor contentt
+				// this should be a variable
+				min-width: 60%;
```
ryanjduffy: Variable for this and the `max-width` below.

LabeledItem/LabeledItem.js
```diff
@@ -9,140 +9,23 @@
...
-
-const Controller = MarqueeController(
-	{marqueeOnFocus: true},
-	Skinnable(
-		ItemBase
-	)
+import Item from '../Item';
+
+// eslint-disable-next-line enact/prop-types
+const LabeledItemBase = ({children, titleIcon, ...rest}) => (
```
ryanjduffy: I forgot we talked about adding warnings so I'll push an update to these shims to address that.

styles/variables-gallium.less
teckliew: Yea 32px is a bit large but it's from here:
itempadding

Item/Item.module.less
Djspaceg: Was this gap visible on a non-retina display? Was scaling set to 100% (or something else)?

styles/variables-gallium.less
ryanjduffy: I see. That seems to be consistent for items other than the menu items.

Item/Item.module.less
ryanjduffy: Doesn't seem to repro at 100% scaling.

LabeledItem/LabeledItem.js
ryanjduffy: Added `deprecate` to each of these.

Item/Item.js
```diff
@@ -27,12 +74,36 @@ import Skinnable from '../Skinnable';
...
-		css={componentCss}
-	/>
-);
+const ItemBase = kind({
+	name: 'Item',
+	styles: {
+		css: componentCss
+	},
+	computed: {
+		className: ({active, selected, styler}) => styler.append({active, selected})
```
ryanjduffy: We may want to consider another prop name than `selected` which overlaps with toggle items like `SwitchItem` and `RadioItem`. It's conceivable we define "selectable items" as different than "toggle items" and implement it by consuming the `selected` prop in toggle items. Or maybe we need to implement this in the base `ToggleItem` and the absence of a "toggle indicator" (e.g. a switch or radio item) we use the border.

What do you think @teckliew and @Djspaceg?

teckliew: We can move both `active` and `selected` to ToggleItem. Although, I'm not sure what the distinction is between `active` and `selected`...
toggleitems

Djspaceg: Does `Item` `selected` differ in purpose/intent from `ToggleItem` `selected`? Curious if we actually need `selected` `itemSelected` and `active`. That seems like a lot of separate states to #1 maintain and #2 for a developer to correctly understand.

Item/Item.module.less
Djspaceg: My opinion would be to ignore this in that case.

Item/Item.module.less
```diff
@@ -16,13 +40,70 @@
...
+			color: @agate-item-selected-color;
+		}
+
 		&.inline {
 			max-width: 240px;
+
+			.itemContent {
+				width: unset;
+
+				&.labelAbove,
```
ryanjduffy: I've added a guard above for `min-width`/`max-width` which allows us to remove this block since those props won't be set for above/below labels.

Djspaceg: Typically what I've done in the past for these sorts of things is include a class on the base component for this "orientation" scenario and get the checking logic out of javascript, then only apply the min/max-width rules to the horizontal state and min/max-height to the vertical state. This means there are no overrides and no logic in JS determining which is ok.

ryanjduffy:
Noting the changes I made:

* Fixed `.label` and `.content` to only apply `max-width` / `min-width` on horizontal layouts
* Added the default `publicClassNames` for item to support overrides for `IconItem`.
* Removed LabeledItem LESS and tests and SlotItem LESS as unused

ryanjduffy:
Also still concerned with the inline items. Right now, here's the result:

image

If we move the `max-width` to the content (rather than the root) and override `display` like this:

```less
	&.inline {
			// overriding ui/Item's inline-block
			display: inline-flex;

			.itemContent {
				max-width: 240px;
			}
	}
```

The result seems better. WDYT?

image

ryanjduffy:
Couple more items:
* Added the inline styles I suggested above based on @teckliew's :+1: in slack. Seems to be an improvement which we could revise later as specs solidify
* Added prop types (but not docs) for Item

PR #139, opened 2019-09-19
PLAT-85707: Fixed Scroller to handle drag events

germboy:
This change allows `Scroller` to disable/enable its spotlight container based on drag events that are emitted.
This change is dependent on another PR: https://github.com/enactjs/enact/pull/2572

### DIFF
--- a/Scrollable/Scrollable.js
+++ b/Scrollable/Scrollable.js
@@ -383,6 +383,16 @@ class ScrollableBase extends Component {
 		);
 	}
 
+	onDragEnd = () => {
+		this.isDragging = false;
+		this.childRef.current.setContainerDisabled(false);
+	}
+
+	onDragStart = () => {
+		this.isDragging = true;
+		this.childRef.current.setContainerDisabled(true);
+	}
+
 	onFlick = ({direction}) => {
 		const bounds = this.uiRef.current.getScrollBounds();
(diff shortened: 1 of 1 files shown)


Scrollable/Scrollable.js
```diff
@@ -340,7 +340,10 @@ class ScrollableBase extends Component {
 		this.stopOverscrollJob('vertical', 'after');
 	}
 
+	flickTarget = null
```
ryanjduffy: We don't appear to be using `flickTarget` other than as a flag that the user flicked. If I'm reading that right, let's rename it here and remove it from the change to `enact`. I'd like to avoid decorating events with additional members so the payloads are consistent. If we need new data, we should create new events.

germboy: Ah, you're right. An earlier iteration actually used the values returned in the event, but I don't believe we need them now.

Scrollable/Scrollable.js
```diff
@@ -697,12 +713,14 @@ class ScrollableBase extends Component {
 	}
 
 	stop = () => {
-		if (!this.props['data-spotlight-container-disabled']) {
+		if (this.isDragging && !this.isFlicked) return;
```
ryanjduffy: I changed this to a bail out to simplify the diff.

PR #140, opened 2019-09-23
ENYO-6274: Remove background color in Panels

ybsung:
Background image was applied to AgateDecorator.
To show it under a panels properly, the panels's background color style should be removed.

I don't know the full history why the background color was added at the first time in https://github.com/enactjs/agate/pull/115. But it seems to me that we need to remove it now.


ybsung:
Closing in favor of https://github.com/enactjs/agate/pull/142.

PR #141, opened 2019-09-23
ENYO-6273: Define letter-spacing and text-index in Icon instead of being inherited from Button

ybsung:

The Icon position in only `Copper` and `Cobalt` skins shifted to the right in Icon Button.

### Resolution

The Icon `letter-spacing` and `text-index` styles were inherited from Button with `normal` value.
But the those styles of Button in the `Copper` and `Cobalt` skins were updated with `9x` in https://github.com/enactjs/agate/pull/97 and https://github.com/enactjs/agate/pull/101.
So we needed to define those styles in Icon separately.
Browser did not recognize `normal` value for them. So I defined those styles in Icon as `initial` to override the those styles in Button.

### Links
ENYO-6273

### DIFF
--- a/styles/variables.less
+++ b/styles/variables.less
@@ -4,5 +4,3 @@
 // App
 // ---------------------------------------
 @agate-app-keepout: 18px;
-@agate-icon-letter-spacing: normal;
-@agate-icon-text-indent: initial;
--- a/styles/variables.less
+++ b/styles/variables.less
@@ -4,3 +4,4 @@
 // App
 // ---------------------------------------
 @agate-app-keepout: 18px;
+@agate-icon-letter-spacing: initial;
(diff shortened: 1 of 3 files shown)


styles/variables.less
```diff
@@ -4,3 +4,5 @@
 // App
 // ---------------------------------------
 @agate-app-keepout: 18px;
+@agate-icon-letter-spacing: normal;
```
Djspaceg: This file is intended as skin-independent global variables.

These two variables should go into `variables-gallium.less`, and it will propagate to all other skins that haven't also set the variable for themselves.

ybsung: I moved them to the `variables-gallium.less`.

Djspaceg: Awesome. Thanks, dude!

PR #142, opened 2019-09-23
ENYO- 6274: Define default panels bg color

ybsung:

The background color of panels in `Copper` skin is white. It should be transparent.

### Resolution

The background color of panels in `gallium day` skin color was defined in https://github.com/enactjs/agate/pull/126/files#diff-09f4b42163b21125118257a2164d1ed5R142
But the `gallium day` skin color was the default color for all skins. So that color was also applied to `Copper` skin. We needed to define it as default background color for all skins separately.

### Links
ENYO-6274

### DIFF
--- a/styles/variables.less
+++ b/styles/variables.less
@@ -4,5 +4,3 @@
 // App
 // ---------------------------------------
 @agate-app-keepout: 18px;
-
-@agate-panels-bg-color: transparent;
--- a/styles/variables.less
+++ b/styles/variables.less
@@ -4,3 +4,5 @@
 // App
 // ---------------------------------------
 @agate-app-keepout: 18px;
+
+@agate-panels-bg-color: transparent;
--- a/styles/colors-carbon.less
+++ b/styles/colors-carbon.less
@@ -135,6 +135,7 @@
 @agate-panels-tab-focus-color:        white;
 @agate-panels-tab-focus-bg-color:     ~"hsla(var(--agate-accent-h, 0), var(--agate-accent-s, 0), var(--agate-accent-l, 0), 0.2)"; // @agate-accent; // fade(@agate-accent, 20%);
 @agate-panels-tab-focus-bg-image:     linear-gradient(to bottom, ~"hsla(var(--agate-accent-h, 0), var(--agate-accent-s, 0), var(--agate-accent-l, 0), 0.1)", @agate-accent 98%, @agate-accent);
+@agate-panels-bg-color:               transparent;
 
 // Picker
 // ---------------------------------------
--- a/styles/colors-cobalt-day.less
+++ b/styles/colors-cobalt-day.less
@@ -145,6 +145,7 @@
 @agate-panels-tab-focus-color:        @agate-foreground-focus;
 @agate-panels-tab-focus-bg-color:     transparent;
 @agate-panels-tab-focus-bg-image:     none;
+@agate-panels-bg-color:               transparent;
 
 // Picker
 // ---------------------------------------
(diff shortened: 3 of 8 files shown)


styles/variables.less
```diff
@@ -4,3 +4,5 @@
 // App
 // ---------------------------------------
 @agate-app-keepout: 18px;
+
+@agate-panels-bg-color: transparent;
```
ryanjduffy: Shouldn't this be in a color file vs the variables file?

Djspaceg: You're absolutely right. This seems different than when I saw it. Maybe I just didn't notice. This definitely belongs in each of the `colors-*.less` files, for the same reasons explained in https://github.com/enactjs/agate/pull/141#discussion_r327225227.

ybsung: I also didn't notice that I had to define color variables in color-*.less files. I fixed it.

ryanjduffy:
While this works as expected, I don't think it's the right change to make the panel background transparent by default. That introduces odd visuals in other contexts like the settings app.

ybsung:
Hello @ryanjduffy 
> That introduces odd visuals in other contexts like the settings app.

Do you mean that the settings app is the `webos-auto-settings` app? The app uses `gallium-night` skin. With and without this PR, the `background-color` of the app panels uses `#1f1f1f`. So I could not understand why that introduces odd visuals?

Djspaceg:
My understanding regarding Panels and Panel background colors and images is that a skin should be able to choose what is right for its visuals. In many cases –Moonstone included– Panels and Panel should be completely translucent. In other cases, Panels could be opaque and Panel would be translucent. This case would produce a special background for Panels and the Panel content would slide across it as transitions happened, like a transparency-sheet or film. Conversely, if Panels was transparent and Panel was opaque, the Panel would slide around the screen like a sheet of paper. Either effect could be desirable according to the design of the skin. I believe this change makes that feature possible in Panels and restores the original styling of the Copper & Cobalt skins.

Please let me know if you feel otherwise so we can discuss.

ryanjduffy:
I think we need to take a position either that Panels has a background color or does not if we want to have predictable results when switching skins. I don't really care which direction we take but I don't think including both will make creating apps with proper skin support possible.

ryanjduffy:
To your point about the interaction panels + panel background, I think that makes sense but this PR does not address that and should. As it is, everything is transparent and looks unreadable for the settings app. My point in this that as a developer of apps, I need predictability for the backgrounds and not have to guess if the bg is going to show through or not.

ybsung:
Hello @ryanjduffy 
Now I got your point. Thanks for sharing your opinion.
But I don't know what to do to get things done. If there is anything that I can do, please let me know.

Djspaceg:
This will likely be corrected as a result of that generalized "base" skin conversion I was talking about recently.

Basically, we set up a new skin (which users can't actually select) that just defines a bunch of default values for all skins, with `0` and `inherit` heavily used for numbers and `transparent` or `inherit` used for colors. This way, as new variables are added, they'll go into the base skin, old less-maintained skins don't break, but also don't get some other skin's values when they may not be appropriate.

I'm currently working on this and will let you know when it's reviewable.

Djspaceg:
Here's the result (PR) of the work referenced above: https://github.com/enactjs/agate/pull/147

ybsung:
Hello @Djspaceg 
Your PR looks great!!! I verified the bg color issue with your branch and it worked well. So we don't need to override the bg color anymore. I'd like to close this PR in favor of https://github.com/enactjs/agate/pull/147.

Djspaceg:
Awyiss! This _should_, at least in theory, address all future cases like this, provided that we establish the rules in this base file correctly and once; so that the contents of it are consistent and trustworthy (since skins will be expecting certain behavior because of them).

PR #143, opened 2019-09-23
PLAT-86338: Add basic shapes to Icon

teckliew:
Add basic shapes (circle, square, triangle) to Icon for general use.


Icon/IconList.js
```diff
@@ -2,6 +2,9 @@
 //
 
 export default {
+	circlelarge       : 0x0EFFB8,  // plus
```
Djspaceg: Comments here should be updated. Perhaps something to the effect of:
```
// literal shape: large circle
// literal shape: square
// literal shape: triangle pointing up
```

Djspaceg:
While some of the icon's comments changed, they don't quite match the request. The names seem to be untouched as well. I think my review must have been unclear.

Switch/Switch.js
```diff
@@ -63,9 +63,9 @@ const SwitchBase = kind({
 			animated: !noAnimation
 		}),
 		children: ({children, skin}) => {
-			if (children !== 'circle') return children;
```
ryanjduffy: If I'm reading this code right, I think we need to refactor this a bit.

If we remove `children` from `defaultProps` (which I know will annoy @webOS101 a bit but seems necessary if the default is contingent on other props), the following will fulfill the requirements that we

* use `children` when set
* use `squarelarge` for "carbon" (since `square` doesn't exist and `stop` is wrong)
* use `circlelage` for everything else

```js
		children: ({children, skin}) => {
			if (children) return children;

			switch (skin) {
				case 'carbon': return 'squarelarge';
				default: return 'circle';
			}
		}
```

Agreed?

webOS101: I'm annoyed that you think I'd be annoyed! :P

Djspaceg: @ryanjduffy your proposal above makes sense to me. I'm in favor of that simplification.

PR #144, opened 2019-09-26
PLAT-85713: Update LabeledIconButton

teckliew:
Update `LabeledIconButton` colors and font to specs.
Related to: https://github.com/enyojs/webos-auto-home/pull/19


Icon/Icon.module.less
```diff
@@ -22,7 +22,7 @@
 		text-indent: @agate-icon-text-indent;
 
 		&.huge {
-			background-position: center -((@agate-icon-sprite-huge-size - @agate-icon-huge-size) / 2);
```
Djspaceg: We'll have to remember this was here, in case we get complaints that the other themes' icons don't work properly.

styles/colors-gallium-day.less
```diff
@@ -126,7 +126,8 @@
 
 // LabeledItem
 // ---------------------------------------
-@agate-labelediconbutton-color: @agate-accent;
+@agate-labelediconbutton-label-color: @agate-white;
```
ryanjduffy: Should be `@agate-text-color`? The text is currently unreadable in sampler for all skins other than gallium-night and other components (e.g. Item) use text color as the unfocused color.

Icon/Icon.module.less
ryanjduffy: @Djspaceg Should this change be applied to the other sizes too?

Djspaceg: Oh! Yes. It should. I didn't notice where this was nested.

styles/colors-gallium-day.less
Djspaceg: We have to be extra careful when making additions/changes to gallium day until the base skin PR (#147) lands. Until then, we need to manually copy the old values from gallium-day into each of the other color files to prevent visually altering them.

ryanjduffy: Sounds like @teckliew should review that first so we can get it landed and then fix up this. Yes?

Djspaceg: Yes. That will help with several new/updated Agate components. However, if it's projected to take a long time to review, longer than it would take to add these values, these could get done and updated later.

Icon/Icon.module.less
teckliew: Changing this to `center center` in the other sizes break the png and svg samples in the sampler.

teckliew: On closer inspection, the assets in sampler are sprites with sizes that won't center the icon with just `center`. It seems like we need to allow custom positioning for non-font icons.

teckliew: On closer inspection, the assets in sampler are sprites with sizes that won't center the icon with just center. It seems like we need to allow custom positioning for non-font icons.

Djspaceg: We could just fix these assets too. It may also be appropriate to have a more holistic approach to this images/sprites support case. One approach could be that we allow a consumer of Icon/IconButton to indicate how many sprites are in their image asset, defaulting to just `1`. This value can be relayed to the LESS via a custom CSS property like `--agate-icon-sprite-count` and we use that to properly calculate the vertical offset of the image in the `background-position` rule. WDYT? Shouldn't actually be difficult, and will work across skins, so images still look correct when a skin is changed.

Djspaceg: @teckliew, I could make this change if you'd prefer.

teckliew: Oh yea go ahead, if you're up for it 👍

Djspaceg: All done! It's now also based on #147 to save some work later.

Docs added, all sizes tested with regular icons, images, sprite images and simplified the legacy junk we no longer need.

Icon/Icon.js
```diff
@@ -58,11 +58,23 @@ const IconBase = kind({
...
-		size: PropTypes.oneOf(['small', 'large', 'huge'])
+		size: PropTypes.oneOf(['small', 'large', 'huge']),
+
+		/**
+		 * The amount of sprite "cells" in the src image.
+		 *
+		 * This property only affects image-based icons, not icon glyphs.
+		 *
+		 * @type {Number}
+		 * @default 0
```
ryanjduffy: There's a disconnect between the documented and actual default below. Which is correct?

Djspaceg: `1` is correct. This should be fixed. `0` should never be used, but if you don't have an image, it doesn't really matter what this value is. I'll update the docs.

PR #147, opened 2019-09-30
Created a new agnostic basic skin

Djspaceg:
tl;dr
Created a new agnostic basic skin.

Also updated several components and existing skins to put math into the skin file rather than the component file(s).

### Intent

The purpose of this change is to find a balance between 1) not breaking existing skins when new features/variables are added and 2) not imposing one skin's values on another skin.

This establishes a base skin that will contain every variable used in any component that is available to a skin. Using and reassigning these variables in a skin is optional. The expectation here is that as new variables are added and assigned to CSS properties that were previously undefined, assigning the base variable value to `inherit` (preferred), `0` or `transparent` should impose the least impact on a skin that should not be affected by these changes or is currently unmaintained.

The intent here is to not impede future development while also not crippling or undermining past development.

### Bonus

Actually, if you're curious to see what this even looks rendered out, simply add `'base'` to the `agate/Skinnable` collection, and add `'base'` to the AgateEnvironment component in Storybook or `skin="base"` to your app instead. It should look like completely plain unstyled content. Boring, but sort of neat I guess. However, remember, this is purely for the reader's enjoyment, not intended to ever be used or seen in a product.


teckliew:
Looks pretty good to me. There some calculations in `Button.less`, `Divider.less`, and `Switch.less`; should we have those in the skin files as well?

Djspaceg:
@teckliew, honestly, we probably should, for consistency and predictability. This does represent a change in how we write the LESS files, but it seems like a generally good approach and separation.

styles/colors-base.less
```diff
@@ -0,0 +1,303 @@
...
+
+// Item
+// ---------------------------------------
+@agate-item-color: inherit;
+@agate-item-focus-bg-color: inherit;
+@agate-item-focus-color: @agate-item-color;
+
+// LabeledItem
+// ---------------------------------------
+@agate-labelediconbutton-color: @agate-text-color;
```
teckliew: `// LabeledIconButton`

Djspaceg:
Regarding Divider, I'm tempted to just leave it alone, because I believe we should deprecate `Divider` and use Heading in its place. It should be a 1:1 conversion, and this would be done during our impending cleanup of Agate. Sound ok? I've corrected Button and Switch.

PR #149, opened 2019-10-03
PLAT-86660: agate/Popup should forward to `onClose` when the close button is tapped/clicked

sugardave:
We want to make sure that `onClose` is called when the popup is closed via the close button.

### DIFF
--- a/Popup/PopupState.js
+++ b/Popup/PopupState.js
@@ -10,7 +10,6 @@ const OpenState = {
 	OPEN: 2
 };
 
-const forwardClose = forward('onClose');
 const forwardHide = forward('onHide');
 
 const PopupState = hoc((config, Wrapped) => {	// eslint-disable-line no-unused-vars
@@ -69,7 +68,6 @@ const PopupState = hoc((config, Wrapped) => {	// eslint-disable-line no-unused-v
 		}
 
 		handlePopupHide = () => {
-			forwardClose(null, this.props);
 			forwardHide(null, this.props);
 
 			this.setState({
--- a/Popup/PopupState.js
+++ b/Popup/PopupState.js
@@ -10,6 +10,7 @@ const OpenState = {
 	OPEN: 2
 };
 
+const forwardClose = forward('onClose');
 const forwardHide = forward('onHide');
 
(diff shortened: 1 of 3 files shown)


Popup/PopupState.js
```diff
@@ -68,6 +69,7 @@ const PopupState = hoc((config, Wrapped) => {	// eslint-disable-line no-unused-v
 		}
 
 		handlePopupHide = () => {
+			forwardClose(null, this.props);
```
ryanjduffy: This is different semantics than we use in moonstone where `onClose` for popup meant that the user requested it close whereas this means that the popup has already closed. Was this intentional?

Djspaceg: Sort of? This may boil down to not having a clear picture, poor heuristics. Do we need different terminology to make this clearer in the future? `handleCloseing` vs `handleClosed`, etc?

ryanjduffy: naming aside, you want `onClose` to be triggered after the popup closes? I'm not clear on that point.

ryanjduffy: and if so, how would an author decide to prevent closing? (that q just occurred to me now)

Djspaceg: Ignoring how this code is currently implemented, and not ignoring naming, honestly `onClose` is vague to me. `onClosing` I would assume to be the event to attach to if I wanted something to fire before the Popup has started its close cycle, as in "upon closing this popup, do X`. Whereas `onClosed`, being past tense, would refer to "now that this action has happened, do Y".

I'm not familiar enough with this legacy code to reconcile the above approach with what it's currently doing. Popup was one of the components we never migrated to `ui`, sadly.

ryanjduffy: For the cases I can recall, our convention in Moonstone has been to emit events (with present tense verbs) when the user takes an action (`onClose` before closing, `onToggle` before toggling). We don't have many notification events but there are some (`onHide` and `onShow` from `ui/Transition` are an example).

Note that I'm only noting a deviation from Moonstone and not stating a direction.

Djspaceg: Present tense for starting an action makes sense to me.

sugardave:
I think this is a better way and easier to follow.  It also keeps the concepts of "hidden" and "open" separated.

PR #151, opened 2019-10-03
PLAT-84609: Add Loading Spinner

teckliew:
Add loading spinner

spinner

### DIFF
--- a/Spinner/Spinner.js
+++ b/Spinner/Spinner.js
@@ -61,7 +61,7 @@ const SpinnerCore = kind({
 	},
 
 	defaultProps: {
-		type: SPINNER_TYPE.searching
+		type: 'searching'
 	},
 
 	styles: {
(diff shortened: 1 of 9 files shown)


Djspaceg:
FYI, this PR will need to include the base-skin initialization variables now that PR #147 has merged.

Spinner/Spinner.js
```diff
@@ -55,7 +61,7 @@ const SpinnerCore = kind({
 	},
 
 	defaultProps: {
-		type: 'searching'
+		type: SPINNER_TYPE.searching
```
Djspaceg: I'm not following the logic behind converting this to a constant object. The type should just be specified as `PropTypes.oneOf('searching', 'loading'), and we can refer to these strings like normal throughout the component. If we're concerned about translations, put the `$L` method in the aria-label area and let everything be simpler.

ryanjduffy:
Closing in favor of #155

PR #152, opened 2019-10-04
ENYO-6276: Forward onClose via clicking on close button if onCloseButtonClick prop is not passed

ybsung:

It seems to me that we don't have to forward `onCloseButtonClick` via clicking out of a popup if `onClose` prop is not passed. I supposed that `onClose`is the main event handler and the `onClose` will be called if the sub event handler is missing like `onCloseButtonClick`.

### DIFF
--- a/Popup/PopupState.js
+++ b/Popup/PopupState.js
@@ -10,6 +10,8 @@ const OpenState = {
 	OPEN: 2
 };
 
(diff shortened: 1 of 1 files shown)


ybsung:
Closing in favor of https://github.com/enactjs/agate/pull/149.

PR #155, opened 2019-10-11
PLAT-84609: Implement Spinner type "loading"

Djspaceg:
Starting from #151, I looked at ways of making the code a little easier to configure (and hopefully maintain). This version leverages fewer LESS variables and automatically and evenly spaces itself given the width of the parent container and an arbitrary amount of circles/dots. As for configuration, you simply set the number of circles, the width of the spinner area, and the duration of the animation. It respects and responds to arbitrary margins and paddings and always remains centered in its space regardless of the height.

### DIFF
--- a/Spinner/Spinner.js
+++ b/Spinner/Spinner.js
@@ -116,7 +116,7 @@ const SpinnerBase = kind({
 		 * The following classes are supported:
 		 *
 		 * * `spinner` - The root component class, unless there is a scrim. The scrim and floating
-		 *   layer can be a sibbling or parent to this root "spinner" element.
+		 *   layer can be a sibling or parent to this root "spinner" element.
 		 *
 		 * @type {Object}
 		 * @public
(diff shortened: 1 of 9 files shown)


Djspaceg:
> Works as advertised. One change; one question below.
> 
> * The animation order appears the opposite from the spec. Spec starts bounce from the left; PR from the right.
> * The spec doesn't call for it it but wondering if there should be a pause between iterations:
> 
> loading

I don't think I understand what you're describing, regarding the animation order. Is the GIF above recorded in sampler using the component? That animation starts from the left (like the spec) and progresses to the right (like the spec).

A pause between is fine, and easy to add, (just rearrange the % values in the animation keyframes); it's just not what was requested.

ryanjduffy:
Sorry ... The GIF above is with the pause and changing the order. Forgot I had changed the order as well

PR #156, opened 2019-10-11
PLAT-87047: Connect Button on the wifi password window is not able to be tapped but must be clicked with mouse device

sugardave:
Advances in our touch event handling for `moonstone` have left `agate` behind.  This patch ensures consistent handling of touch and pointer events for `agate/Input`.

### DIFF
--- a/Input/pointer.js
+++ b/Input/pointer.js
@@ -21,11 +21,23 @@ const handlePointerUp = handle(
 	stop						// (and stop propagation to support touch)
 );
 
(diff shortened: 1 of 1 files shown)


Djspaceg:
This does fix this issue on the Auto hardware/touchscreen! The UX is a little funny because you have to tap to dismiss the VKB, then tap connect again, but maybe that's ok? (Not the fault of this PR.)

PR #157, opened 2019-10-16
PLAT-87182: Add support for in-theme localization...

JayCanuck:
...using updated ResBundle basepath format.

* Adds an `internal/$L.js` and theme-level `./resources` directory, similar to Moonstone and the theme template.

Enact-DCO-1.0-Signed-off-by: Jason Robitaille <jason.robitaille@lge.com>


internal/$L/$L.js
```diff
@@ -0,0 +1,100 @@
...
+ * @returns {Promise|ResBundle} Resolves with a new ilib.ResBundle
+ */
+function createResBundle (options) {
+	let opts = options;
+
+	if (typeof ILIB_AGATE_PATH !== 'undefined') {
+		opts = {
+			loadParams: {
+				// Deprecated; to be removed in future
+				root: ILIB_AGATE_PATH
```
ryanjduffy: Since we don't have any legacy ilib support for agate, let's start off clean without this fallback.

JayCanuck: Updated

PR #159, opened 2019-10-18
PLAT-85965: Support for IconButton Spinner

teckliew:
- Add `size="huge"` to `Spinner` with additional tweaks to the spinner size.
- Add support for `IconButton` with spinners
spinnerloader

Usage:
```
// normal IconButton usage:
<IconButton>home</IconButton>

//IconButton with Spinner
<IconButton>
  <Spinner />
<IconButton>
```


styles/variables-carbon.less
```diff
@@ -319,18 +319,29 @@
 // ---------------------------------------
 @agate-spinner-size: 52px;
 @agate-spinner-bg-padding: 9px;
+@agate-spinner-huge-size: (@agate-spinner-size * 2.5);
+@agate-spinner-huge-bg-padding: (@agate-spinner-bg-padding * 2.5);
 @agate-spinner-small-size: 36px;
 @agate-spinner-small-bg-padding: 5px;
+@agate-spinner-smallest-size: 24px;
 @agate-spinner-searching-width: (@agate-spinner-size / 13);
```
Djspaceg: I'm still confused by the `/ 13` value here and the `* 3` below. Perhaps this should just be set to `4px` and the below value set to `12px`? It seems like the fixed values would be easier to mentally process and update (if needed) rather than fifty-two thirteeths of the width.

styles/variables-base.less
```diff
@@ -363,19 +363,33 @@
 @agate-spinner-size: inherit;
 @agate-spinner-bg-margin: inherit;
 @agate-spinner-bg-padding: inherit;
+@agate-spinner-huge-size: inherit;
```
Djspaceg: It may be a good idea to start adding these variables' relationships and default sizes into this file now, to save a bit of work here and in the future. Thoughts?

styles/variables-carbon.less
```diff
@@ -319,18 +319,29 @@
 // ---------------------------------------
 @agate-spinner-size: 52px;
 @agate-spinner-bg-padding: 9px;
+@agate-spinner-huge-size: (@agate-spinner-size * 2.5);
```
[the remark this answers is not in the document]
teckliew: based on the ratio with `IconButton`'s normal and huge sizes. Should we just set it as a value like `130px` instead?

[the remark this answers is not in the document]
Djspaceg: Yes. I believe that will be easier to understand. Imagine the scenario where we get new requirements and the normal spinner has to be 6px smaller. We didn't get any instruction to change the huge spinner, but now, because there's a relationship to the size of the normal spinner, now huge spinner gets smaller, if we don't notice the relationship.

teckliew:
After shifting things around, I opened up `iconComponent` as a prop so that now `IconButton` is more simplified. We can just plug in `Spinner` as a child.

IconButton/IconButton.module.less
```diff
@@ -13,6 +13,7 @@
 		box-sizing: border-box;
 		font-size: @agate-iconbutton-font-size;
 		vertical-align: middle;
+		text-indent: 0;
```
Djspaceg: Let's make sure to not lose this with the deprecation of IconButton in #169. This is important for keeping the icon centered inside the Button.

Spinner/Spinner.js
```diff
@@ -129,11 +129,11 @@ const SpinnerBase = kind({
 		 * Recommended usage is "medium" (default) for standalone and popup scenarios, while "small"
 		 * is best suited for use inside other elements, like {@link agate/SlotItem.SlotItem}.
 		 *
-		 * @type {('medium'|'small')}
+		 * @type {('huge'|'medium'|'small')}
 		 * @default 'medium'
 		 * @public
 		 */
-		size: PropTypes.oneOf(['medium', 'small']),
+		size: PropTypes.oneOf(['huge', 'medium', 'small']),
```
Djspaceg: We'll have to support `'smallest'` here too

Djspaceg:
Usage changed a bit from the initial version of this, with the demise of `IconButton`. Here's the currently supported approach:

```
<Button iconComponent={Spinner} icon=" " />
```

The `icon=" "` is currently a workaround for the logic in `ui` which unnecessarily expects icon to be assigned to a truthy string. It needs to be updated to support booleans as well, so you could simply put `<Button iconComponent={Spinner} icon />` or something like that. Open to ideas on this implementation as I'm not settled on this _cleverness_, unsure of it being intuitive enough.

If you wanted the loading spinner you can easily enough create a proxy method:
```
const Loader = (props) => <Spinner {...props} type="loading" />;

// and then do

<Button iconComponent={Loader} icon=" " />
```

PR #160, opened 2019-10-21
PLAT-85937: Add deep-freeze in dev mode

ryanjduffy:
Prior implementations (#134, #135) were naive and should have used `hasOwnProperty()`. Instead of adding that, added the small `deep-freeze` module which handles this correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "@enact/ui": "^3.1.2",
     "classnames": "^2.2.6",
     "color-convert": "^1.9.3",
+    "deep-freeze": "0.0.1",
     "eases": "^1.0.8",
     "immer": "^3.2.0",
     "prop-types": "^15.6.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -32,6 +32,7 @@
     "@enact/ui": "^3.1.2",
     "classnames": "^2.2.6",
     "color-convert": "^1.9.3",
+    "deep-freeze": "0.0.1",
```
Djspaceg: Should this be a dev-dependency?

ryanjduffy: No. `devDependencies` are for build-time or maintainer dependencies. This is a regular dependency even though it only applies to "dev mode"

ryanjduffy: More concretely, if it were a dev dependency, if an author npm installed agate and tried to `enact serve` their app, it would throw because `deep-freeze` was missing.

JayCanuck: Makes sense as a regular dependency since it's content being included in builds, rather than a build/test tool :+1:

PR #161, opened 2019-10-21
PLAT-86999: Suppress onToggle when disabled

ryanjduffy:
## Issue
Tapping a SwitchItem when disable will still emit the `onToggle` event

## Resolution
`onToggle` is mapped to `onClick` by `Toggleable` so adding a handler that suppresses forwarding `onClick` when disabled resolves the issue.

### DIFF
--- a/SwitchItem/SwitchItem.js
+++ b/SwitchItem/SwitchItem.js
@@ -1,7 +1,7 @@
 /**
- * Provides Agate-ez-themed Item component and interactive toggle switch icon.
+ * Provides Agate-themed Item component and interactive toggle switch icon.
  *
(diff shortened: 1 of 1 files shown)


SwitchItem/SwitchItem.js
```diff
@@ -88,6 +98,7 @@ const SwitchItemBase = kind({
 	},
 
 	defaultProps: {
+		disabled: false,
 		offText: 'off',
```
Djspaceg: Think we should `$L` these strings since we're in here already?

webOS101: They wouldn't get updated on locale change if we did it here.

PR #162, opened 2019-10-29
PLAT-88681: Update with new icons

teckliew:
Add 23 new icons and update 6 icons.
newicons


webOS101:
Let's update this before we do our next release or on next font update.  As this is never installed on systems (yet) we can use agate release # if we need to disambiguate.

PR #163, opened 2019-10-30
PLAT-86995: Set Base Variables

teckliew:
Consolidated less variables to the base skin.
Used variables that were most used into base which eliminates duplicate varaibles across other skins; this ends up with using a lot of styles from `copper`, `cobalt`, and `titanium`.


styles/variables-base.less
```diff
@@ -239,54 +248,54 @@
...
 @agate-panels-tab-margin: inherit;
 @agate-panels-tab-padding: inherit;
-@agate-panels-tab-separator-border-width: inherit;
-@agate-panels-vertical-tab-after-padding-right: inherit;
-@agate-panels-vertical-tab-before-right: inherit;
-@agate-panels-vertical-tab-bg-position: inherit;
-@agate-panels-vertical-tab-group-margin: inherit;
-@agate-panels-vertical-tab-group-padding: inherit;
-@agate-panels-vertical-tab-margin: inherit;
+@agate-panels-tab-separator-border-width: 1px;
```
Djspaceg: This one seems like a skin-specific candidate. Let's set this to `0` here and `1px` in the skins where it's not zero.

styles/variables-base.less
```diff
@@ -239,54 +248,54 @@
...
-@agate-picker-border-radius: inherit;
-@agate-picker-transform: inherit;
-@agate-picker-label-transform: inherit;
-@agate-picker-font-weight: inherit;
-@agate-picker-font-size: inherit;
-@agate-picker-active-font-size: inherit;
+@agate-picker-border-radius: 0;
+@agate-picker-transform: none;
+@agate-picker-label-transform: none;
+@agate-picker-font-weight: 500;
```
Djspaceg: Skin-specific, let's set this to the `@agate-base-font-weight` variable and update the skins as needed.

styles/variables-base.less
```diff
@@ -296,108 +305,107 @@
...
+@agate-slider-vertical-padding: (@agate-slider-height / 2) ((@agate-slider-height - @agate-slider-track-height) / 2);
 @agate-slider-height: @agate-button-height;
 @agate-slider-width: @agate-slider-height;
-@agate-slider-track-height: inherit;
+@agate-slider-track-height: 12px;
 @agate-slider-track-width: @agate-slider-track-height;
-@agate-translate-center: inherit;
-@agate-slider-bg-transform: inherit;
-@agate-slider-track-transform: inherit;
+@agate-translate-center: translate3d(-50%, -50%, 0);
```
Djspaceg: Let's move this up to the top section, around the "base" or "keepout" area, since it's more of a general utility variable.

styles/variables-base.less
```diff
@@ -296,108 +305,107 @@
...
+@agate-incrementslider-border-radius: 0;
 @agate-incrementslider-horizontal-margin: @agate-slider-horizontal-margin;
-@agate-incrementslider-horizontal-slider-margin: inherit;
-@agate-incrementslider-horizontal-slider-padding: inherit;
-@agate-incrementslider-horizontal-button-margin: inherit;
-@agate-incrementslider-horizontal-button-decrement-border-radius: inherit;
-@agate-incrementslider-horizontal-button-decrement-border-width: inherit;
-@agate-incrementslider-horizontal-button-increment-border-radius: inherit;
-@agate-incrementslider-horizontal-button-increment-border-width: inherit;
+@agate-incrementslider-horizontal-slider-margin: 0 -1px;
```
Djspaceg: This `-1px` seems skin-specific. `0` should be sufficient for the base value, and skins that really need a negative 1 horizontal margin are welcome to set one, but I get the impression that this is a pretty specific visual.

styles/variables-base.less
```diff
@@ -296,108 +305,107 @@
...
+@agate-incrementslider-vertical-button-decrement-border-width: 0;
+@agate-incrementslider-vertical-button-increment-border-radius: 0;
+@agate-incrementslider-vertical-button-increment-border-width: 0;
 
 // RadioItem
 // ---------------------------------------
 @agate-radioitem-separator-border-width: @agate-divider-border-width;
 @agate-radioitem-icon-border-width: @agate-button-border-width;
-@agate-radioitem-icon-border-radius: inherit;
+@agate-radioitem-icon-border-radius: 999px;
```
Djspaceg: How do you feel about setting a global skin corner radius variable and then using it throughout this file. That way, a skin could set the variable once and affect everything that is based on that border-radius uniformly. Ex: Electro rarely uses a fully round component, and could simply set this one variable once and appect all related components quickly.

styles/variables-base.less
```diff
@@ -296,108 +305,107 @@
...
-@agate-switch-selected-icon-position-left: inherit;
+@agate-switch-icon-font-size: (@agate-switch-height * @agate-switch-icon-scale);
+@agate-switch-icon-position-left: 0;
+@agate-switch-selected-icon-position-left: (@agate-switch-width - @agate-switch-height);
 
 // ToggleButton
 // ---------------------------------------
-@agate-togglebutton-indicator-position: inherit;
-@agate-togglebutton-indicator-border-width: inherit;
+@agate-togglebutton-indicator-position: auto 21px 9px 21px;
```
Djspaceg: It seems odd, but this is probably best to set to something like `0 0 0 0` so a skin is encouraged to use its own positioning.

styles/variables-base.less
```diff
@@ -296,108 +305,107 @@
...
+@agate-togglebutton-indicator-border-width: 1px;
```
Djspaceg: I'm hesitent to assign specific border-widths to components since it's such a stylistic choice. If this is 0 by default, I think that will work out better for us.

styles/variables.less
```diff
@@ -3,4 +3,3 @@
 
 // App
 // ---------------------------------------
-@agate-app-keepout: 18px;
```
Djspaceg: This removal may actually cause problems, since some apps import this file and refer to this variable.

ryanjduffy:
I added the following to try to compile with only the base rules and found 4 missing base variables:

```less
	.buildSkinRules(base; {
		// Load the skin rules into this scope
		@componentRules();
	});
```

```less
@agate-disabled-opacity
@agate-item-label-color
@agate-item-selected-color
@agate-iconbutton-selected-border-color
```

Djspaceg:
We may want to include that above code in the skin.less file to help validate this in the future.

Djspaceg:
@ryanjduffy We'll need to document it if we don't fix it right now. I don't have a big problem with the out-of-scope issue, just so it can get resolved in a timely manner for the next release.

I'm not sure what you mean by "invalid defaults". `border-radius: inherit;` is definitely valid. Maybe that was an unfortunate example? The only time I've seen where the setting is invalid is for `border-color` where CSS is actually inheriting `color` and not the parent's `border-color` or when we're doing math on a variable and we've set it to `inherit`, which obviously isn't a number and crashes the computation.

ryanjduffy:
You're right. That was the wrong example. `border-color` was the example:

image

ryanjduffy:
Another example was `text-indent` for Button which is reusing `@agate-button-letter-spacing` for the value that defaults to `normal` (which is valid for `letter-spacing` but not `text-indent`).

Djspaceg:
> Another example was `text-indent` for Button which is reusing `@agate-button-letter-spacing` for the value that defaults to `normal` (which is valid for `letter-spacing` but not `text-indent`).

Correct. We should likely sever the relationship between those variables in the base file to avoid issues. In skins where the text-indent must be offset by the letter-spacing, we should establish the relationship between these two variables, but only in those skins. This way, the `normal` value can be used in `letter-spacing` for "regular" skins and `text-align: justify` will still work (unlike if we simply assigned `0`).

PR #166, opened 2019-11-05
PLAT-88733: Replace Divider with Heading

RenukaAtale:
- Remove Divider files from Agate


Drawer/Drawer.js
```diff
@@ -69,10 +69,10 @@ const DrawerBase = kind({
 });
 
 /**
- * Applies Agate specific behaviors to [DividerBase]{@link agate/Divider.DividerBase}.
```
sugardave: These should change to `DrawerBase` and `Drawer`

Popup/Popup.js
```diff
@@ -81,10 +81,10 @@ const PopupBase = kind({
 });
 
 /**
- * Applies Agate specific behaviors to [DividerBase]{@link agate/Divider.DividerBase}.
```
sugardave: These should change to `PopupBase` and `Popup`

RadioItem/RadioItem.js
```diff
@@ -33,7 +33,7 @@ import componentCss from './RadioItem.module.less';
 const RadioItemBase = kind({
 	name: 'RadioItemBase',
 
-	propTypes: /** @lends agate/Divider.DividerBase.prototype */ {
```
sugardave: These should change to `RadioItemBase` and `RadioItem`

SwitchItem/SwitchItem.js
```diff
@@ -35,7 +35,7 @@ const Switch = Skinnable({prop: 'skin'}, SwitchBase);
 const SwitchItemBase = kind({
 	name: 'SwitchItemBase',
 
-	propTypes: /** @lends agate/Divider.DividerBase.prototype */ {
```
sugardave: These should change to `SwitchItemBase` and `SwitchItem`

styles/colors-carbon.less
```diff
@@ -169,7 +170,7 @@
 // RadioItem
 // ---------------------------------------
 @agate-radioitem-focus-color: @agate-item-color;
-@agate-radioitem-separator-border-color: @agate-divider-border-color;
+@agate-radioitem-separator-border-color: fadeout(@agate-text-color, 30%);
```
sugardave: I think this should be `@agate-heading-border-color`
(and in all the other `.less` files, this change should be made)

styles/colors-gallium-day.less
```diff
@@ -66,11 +66,6 @@
 @agate-colorpicker-bg-color: rgba(0, 0, 0, 0.5);
 @agate-colorpicker-palette-shadow: @agate-popup-shadow;
 
-// Divider
-// ---------------------------------------
-@agate-divider-color:        @agate-text-color;
```
sugardave: It doesn't look like there is a new `@agate-heading-color` here, also there are 2 definitions for `@agate-heading-border-color` below (line 83 and 86)

Same thing in `colors-gallium-night.less`

styles/variables-base.less
```diff
@@ -84,18 +84,6 @@
 @agate-datetimepicker-label-transform: inherit;
 @agate-datetimepicker-cell-transform: inherit;
 
-// Divider
-// ---------------------------------------
-@agate-divider-border-width: inherit;
-@agate-divider-font-size: inherit;
-@agate-divider-font-weight: inherit;
-@agate-divider-height: inherit;
-@agate-divider-spacing: @agate-component-spacing;
```
sugardave: Should we have `Heading` replacements for `@agate-divider-spacing` and `@agate-divider-font-ligatures`?

styles/colors-carbon.less
Djspaceg: @sugardave is correct. Setting this to `@agate-heading-border-color` preserves the relationships between the variables rather than assigning two separate values which happen to be the same value.

styles/variables-base.less
RenukaAtale: `@agate-divider-spacing` is already present in Heading as `@agate-heading-spacing-medium`
We don't need that anymore `@agate-divider-font-ligatures`

PR #168, opened 2019-11-15
PLAT-84902: Create LabeledSlider

ybsung:
I created `LabeledSlider` for Settings App and Volume App.
The `LabeledSlider` allows `decrementIcon` and `incrementIcon` props for icons and `decrementText` and `incrementText` props for texts. In addition, the slider could have icon or text elements as `children` prop instead of those props so app developers to customize styles for the icons or the texts.

### Links
PLAT-84902

### Comments

GUI problem
When dragging the knob of LabeledSlider to the rightmost or the leftmost, then the knob positions over the icons or the texts. This is the GUI problem. I'm still discussing it with the GUI developer. I'll fix it in another seperate issue.

Additional work
I removed `IncrementSlider/utils.js` which might not be used anywhere.
I fixed the lint for agate/ProgressBar.


ybsung:
No more available.

PR #169, opened 2019-11-16
PLAT-88683 Iconbutton needs to be reconciled with Button with Icon component

RenukaAtale:
- Add new sizes to the button component ( smallest and huge)
- Add new sizes to the Icon component ( smallest and huge)
- Reconciled IconButton component into Button component
- Updated the style for the new sizes

### DIFF
--- a/Button/Button.js
+++ b/Button/Button.js
@@ -122,7 +122,7 @@ const ButtonBase = kind({
 	},
 
 	computed: {
-		className: ({backgroundOpacity, children, highlighted, joinedPosition, selected, type, styler, size}) => styler.append(
+		className: ({backgroundOpacity, highlighted, joinedPosition, selected, type, styler, size}) => styler.append(
 			backgroundOpacity,
 			size,
 			type,
--- a/Button/Button.js
+++ b/Button/Button.js
@@ -93,19 +93,11 @@ const ButtonBase = kind({
 		/**
 		 * The size of the button.
 		 *
-		 * @type {('smallest', 'small', 'large', 'huge')}
+		 * @type {('smallest'|'small'|'large'|'huge')}
 		 * @default 'large'
 		 * @public
 		 */
 		size: PropTypes.oneOf(['smallest', 'small', 'large', 'huge']),
-
-		/**
-		 * The amount of sprite "cells" in the src image.
-		 *
-		 * @type {Number}
-		 * @public
-		 */
-		spriteCount: PropTypes.number,
 
 		/**
 		 * Specify how this button will be used. Is it a standalone button, or is it in a grid of
@@ -137,11 +129,10 @@ const ButtonBase = kind({
 			{
 				highlighted,
 				selected,
-				noChildren: (Boolean(!children))
+				noChildren: (!Boolean(children))
 			}
 		),
 		minWidth: ({children}) => (!children)
-
 	},
 
 	render: ({css, ...rest}) => {
--- a/Button/Button.js
+++ b/Button/Button.js
@@ -16,6 +16,7 @@ import Spottable from '@enact/spotlight/Spottable';
 import {ButtonBase as UiButtonBase, ButtonDecorator as UiButtonDecorator} from '@enact/ui/Button';
 import Pure from '@enact/ui/internal/Pure';
 import PropTypes from 'prop-types';
+import React from 'react';
 import compose from 'ramda/src/compose';
 
 import {IconBase} from '../Icon';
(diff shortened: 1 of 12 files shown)


styles/variables-base.less
```diff
@@ -51,22 +51,31 @@
...
 @agate-button-highlighted-opacity: 1;
 @agate-button-skew: 0;
 @agate-button-grid-border-radius: 0;
 @agate-button-bg-transform: none;
+@agate-button-icon-border-radius: 9999px;
+@agate-button-smallest-height: 48px;
+@agate-button-smallest-font-size: 15px;
+@agate-button-smallest-line-height: (@agate-button-smallest-height - (2 * @agate-button-border-width));
+@agate-button-smallest-h-padding: 15px;
+@agate-button-small-height: 63px;
```
Djspaceg: This appears to be 3px larger. Why?

styles/variables-base.less
```diff
@@ -51,22 +51,31 @@
...
+@agate-button-small-font-size: 33px;
```
Djspaceg: This appears to be 9px larger. Why?

styles/variables-carbon.less
```diff
@@ -11,6 +11,8 @@
 // ---------------------------------------
 @agate-button-border-width: 1px;
 @agate-button-border-radius: 9999px;
+@agate-button-font-weight: 400;
+
```
Djspaceg: Extra line

styles/variables-base.less
```diff
@@ -166,19 +175,22 @@
...
 @agate-icon-margin: @agate-component-spacing;
 @agate-icon-letter-spacing: inherit;
 @agate-icon-text-indent: initial;
+@agate-icon-smallest-size: 21px;
+@agate-icon-smallest-font-size: (@agate-icon-smallest-size * 2);
+@agate-icon-sprite-smallest-size: @agate-icon-smallest-size;
+@agate-icon-small-size: 36px;
+@agate-icon-small-font-size: (@agate-icon-small-size * 2);
+@agate-icon-sprite-small-size: @agate-icon-small-size;
+@agate-icon-button-small-size: @agate-button-small-height;
```
Djspaceg: Do these variables: `@agate-icon-button-small-size` make sense to retain or should they be renamed to fit into the Button section? Where are they used?

styles/variables-cobalt.less
```diff
@@ -12,12 +12,11 @@
 
 // Button
 // ---------------------------------------
+@agate-button-border-radius: 9999px;
+@agate-button-font-size: 27px;
 @agate-button-font-weight: 700;
 @agate-button-letter-spacing: 9px;
-@agate-button-text-indent: @agate-button-letter-spacing;
```
Djspaceg: I don't think this rule can be removed for this skin.

styles/variables-copper.less
```diff
@@ -12,12 +12,10 @@
 
 // Button
 // ---------------------------------------
+@agate-button-font-size: 27px;
 @agate-button-font-weight: 700;
 @agate-button-letter-spacing: 9px;
-@agate-button-text-indent: @agate-button-letter-spacing;
```
Djspaceg: I don't think this rule can be removed for this skin.

styles/variables-gallium.less
```diff
@@ -43,22 +44,13 @@
 @agate-heading-title-decoration-opacity: 0.8;
 @agate-heading-title-decoration-transform: skew(-10deg);
 
-// Icon Sizes
-// ---------------------------------------
-@agate-icon-small-size: 28px;
-@agate-icon-huge-font-size: 400px;
```
Djspaceg: I think removing this drastically changes the size of the huge icon in Gallium. It would go from 400px to a little over 200px. Is this intentional?

styles/variables-titanium.less
```diff
@@ -11,6 +11,8 @@
 // Button
 // ---------------------------------------
 @agate-button-border-radius: 9999px;
+@agate-button-font-weight: 400;
+
```
Djspaceg: Extra line

Button/Button.js
```diff
@@ -121,13 +122,17 @@ const ButtonBase = kind({
...
 			type,
 			(joinedPosition && 'joined' + cap(joinedPosition)),  // If `joinedPosition` is present, prepend the word "joined" to the variable, so the classes are clearer.
-			{highlighted, selected}
+			{
+				highlighted,
+				selected
+			}
 		),
-		minWidth: ({children}) => (!children)
+		minWidth: ({children}) => (React.Children.count(children) === 0)
```
webOS101: This computation is not exactly the same.  Passing an empty string `''` will yield one child but would previously have set `minWidth`.  What was the motivation for this change?

Djspaceg: I'd added that. I thought it would do a better job at catching empty children.

webOS101: Is this change in function acceptable or should we revert it?

webOS101: Alternately...  Add `|| children === ''`?

PR #170, opened 2019-11-20
PLAT-85962: Add PopupMenu Support for Scrolling

teckliew:
Add scrolling to `PopupMenu`, allowing scrollable elements in popup.

### DIFF
--- a/PopupMenu/PopupMenu.js
+++ b/PopupMenu/PopupMenu.js
@@ -9,6 +9,7 @@ import Transition from '@enact/ui/Transition';
 import Skinnable from '../Skinnable';
 import Heading from '../Heading';
 import LabeledIconButton from '../LabeledIconButton';
+import Scroller from '../Scroller';
 
 import PopupState from '../Popup/PopupState';
 
@@ -19,6 +20,7 @@ const PopupMenuBase = kind({
 	propTypes: {
 		closeButton: PropTypes.bool,
 		css: PropTypes.object,
+		orientation: PropTypes.string,
 		noAnimation: PropTypes.bool,
 		onClose: PropTypes.func,
 		onHide: PropTypes.func,
@@ -37,7 +39,11 @@ const PopupMenuBase = kind({
 		className: 'popupMenu'
 	},
 
-	render: ({children, closeButton, css, noAnimation, onClose, onHide, open, title, ...rest}) => {
+	computed: {
+		bodyClassName: ({css, orientation}) => orientation === 'horizontal' ? css.horizontalBody : css.body
+	},
+
+	render: ({bodyClassName, children, closeButton, css, noAnimation, onClose, onHide, open, orientation, title, ...rest}) => {
 		return (
 			<Transition
 				noAnimation={noAnimation}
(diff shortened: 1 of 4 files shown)


styles/variables-gallium.less
```diff
@@ -108,7 +108,7 @@
 
 // PopupMenu
 // ---------------------------------------
-@agate-popupmenu-title-margin-bottom: 136px;
+@agate-popupmenu-title-margin: 136px;
```
Djspaceg: To match with the previous version of the implementation, shouldn't this be `0 0 136px 0`?

PopupMenu/PopupMenu.js
```diff
@@ -37,7 +39,11 @@ const PopupMenuBase = kind({
 		className: 'popupMenu'
 	},
 
-	render: ({children, closeButton, css, noAnimation, onClose, onHide, open, title, ...rest}) => {
+	computed: {
+		bodyClassName: ({css, orientation}) => orientation === 'horizontal' ? css.horizontalBody : css.body
```
Djspaceg: How do you feel about having this class just be `.horizontal` and then we can use a `.horizontal .body` selector to target the necessary no-wrap rule. This would mean changing this prop to `className` from `bodyClassName`.

Also, we could set `orientation` to default to "horizontal" so the other PR isn't needed: https://github.com/enyojs/webos-auto-home/pull/38/files#diff-32b7ef32f74375c7a9715daa2e883fd7R60

PopupMenu/PopupMenu.js
```diff
@@ -53,7 +59,7 @@ const PopupMenuBase = kind({
 					<Cell className={css.title} shrink>
 						<Heading size="title">{title}</Heading>
 					</Cell>
-					<Cell className={css.body} shrink>
```
Djspaceg: Was the `shrink` prop not cooperating with the new `Scroller`?

PopupMenu/PopupMenu.module.less
```diff
@@ -11,7 +11,11 @@
 		background-image: @agate-popupmenu-bg-image;
 
 		.title {
-			margin-bottom: @agate-popupmenu-title-margin-bottom;
+			margin: @agate-popupmenu-title-margin;
+		}
+
+		.horizontalBody {
```
Djspaceg: This would become `.horizontal .body {` based on the above comment, and `className={css.body}` just just remain how it was.

PopupMenu/PopupMenu.js
```diff
@@ -23,21 +24,27 @@ const PopupMenuBase = kind({
 		onClose: PropTypes.func,
 		onHide: PropTypes.func,
 		open: PropTypes.bool,
+		orientation: PropTypes.oneOf(['horizontal']),
```
Djspaceg: The idea here is that we technically, at this time, only support one orientation, but want to leave it open to future implementations that allow for vertical scrolling (in the code, and in an updated proptypes), so we only actually list the one that's supported presently. This also directly injects into the className field, so we want to ensure that the prop, if provided is spelled correctly and will work as expected.

PopupMenu/PopupMenu.js
Djspaceg: `shrink` was originally removed because it was not cooperating with Scroller's auto-sizing. Moving Scroller into the Cell as a child allowed to Scroller to set itself to the correct dimensions, and then Cell could shrink-wrap around Scroller so they can both be positioned automatially by flex-box.

PR #172, opened 2019-11-20
PLAT-88684 Button needs to be able to animate its separate pieces

RenukaAtale:
Update Button component to support animation.
- Add 2 props to the Button component 
  Animation : is `string` prop  will be applied on `onRender`. 
  AnimationDelay:  is the `number` prop which will be used to apply animation on buttons group.

- Added css classes for animation on button


Button/Button.js
```diff
@@ -43,6 +43,24 @@ const ButtonBase = kind({
 	name: 'Button',
 
 	propTypes: /** @lends agate/Button.ButtonBase.prototype */ {
+		/**
+		 * Provides a way to apply animation on this button once the component is rendered. 
+		 *
+		 * @type {String}
```
webOS101: This should use the same format as other `oneOf`.  `{('onRender')}`

Button/Button.module.less
```diff
@@ -5,178 +5,210 @@
...
-		box-sizing: border-box;
-		font-size: @agate-button-font-size;
-		vertical-align: middle;
-
-		&.minWidth {
-			padding: 0;
-
-			.client .icon {
-				margin: 0;
+	.applySkins( {
```
webOS101: Can you remove this extra space?  Also, the indentation below seems one level too many.

Button/Button.module.less
```diff
@@ -5,178 +5,210 @@
...
+	);
 
 	// TODO: Replace with an updated .focus() mixin that can apply rules both when focused and
 	// within a focused node
 	&:global(.spottable):focus,
 	:global(.spottable):focus & {
-		.applySkins({
-			background-color: transparent;
-			background-image: none;
+		.applySkins( {
```
webOS101: Same space and indentation issue here.

Djspaceg:
I'm working on resolving the merge conflicts.

Djspaceg:
Below is some example code to verify that the animation effect is working as as intended:
```
<div style={{padding: '3em'}}>
	<Button animate="onRender">A</Button>
	<Button animate="onRender" animationDelay={100}>B</Button>
	<Button animate="onRender" animationDelay={200}>C</Button>
	<Button animate="onRender" animationDelay={300}>D</Button>
</div>
```

Button/Button.module.less
```diff
@@ -4,6 +4,16 @@
 @import "../styles/variables.less";
 @import "../styles/skin.less";
 
+@keyframes scale {
```
ryanjduffy: Where is this used? I'm missing the reference

Button/Button.module.less
```diff
@@ -4,6 +4,16 @@
...
+	0%   { transform: scale(0); }
+	100% { transform: scale(1); }
+}
+@keyframes scale-bounce {
```
ryanjduffy: This animation conflicts with the styling of buttons in the electro skin.

@Djspaceg - Do you want to address that?

Button/Button.module.less
Djspaceg: It isn't used. Was considering using in other skins or in an animation bank. It doesn't technically need to be in this PR.

Button/Button.module.less
Djspaceg: I'll look into this real quick.

Djspaceg: This will require a much more elaborate animation system, with animations that are specific to a skin, which isn't currently implemented. `@keyframe` sequences can't be wrapped by selectors, and therefore fall outside the realm of skin's reach. We need a system that provides a global cache of skin-name-prefixed animations which can reference skin variables to be able to support this and a proper animation bank.

FWIW, the following is what the keyframes will look like to properly implement this across all skins:
```
@keyframes scale-bounce {
	0%   { transform: scale(0); animation-timing-function: ease-out; }
	80%  { transform: scale(1.07); animation-timing-function: ease-in-out; }
	100% { transform: scale(1); }
}

@keyframes scale-bounce-bg {
	0%   { transform: @agate-button-bg-transform, scale(0); animation-timing-function: ease-out; }
	80%  { transform: @agate-button-bg-transform, scale(1.07); animation-timing-function: ease-in-out; }
	100% { transform: @agate-button-bg-transform, scale(1); }
}
```
I was working on a system to support this, but had to back-burner it for today's work priority.

ryanjduffy: For clarity, you'd like to move ahead with this PR then?

Djspaceg: I think so. I don't have a better solution. I could include a note about what you noticed and why we "chose what we chose" for this case.

ryanjduffy: This trail is sufficient I think.

PR #175, opened 2019-11-23
PLAT-95919: Add travis task and fix some linting and test failures

webOS101:
Adds a Travis task to run tests and linting.  However, there are a lot lint warnings and some failing tests.


Slider/tests/Slider-specs.js
```diff
@@ -279,7 +279,7 @@ describe('Slider', () => {
 		expect(actual).toBe(expected);
 	});
 
-	test('should not call onSpotlightLeft on horizontal slider at greater than min value', () => {
+	test.skip('should not call onSpotlightLeft on horizontal slider at greater than min value', () => {
```
webOS101: Should confirm with Blake but I believe we didn't pull this spotlight/5-way code over from Moonstone so these can probably be deleted.

PR #176, opened 2019-11-26
PLAT-85942: Add badge support to Button

Djspaceg:
Added notification bubble (badge) support to `Button`.

The badge is customizable at the skin level and supports a custom color (like the `Heading` decoration element).


ryanjduffy:
Forgot to mention that focus styling may need to be revisted as well.

image

Djspaceg:
> Forgot to mention that focus styling may need to be revisted as well.
> 
> image

It's funny you mention this, because I also had noticed and been concerned about this. The design doesn't include a depiction of the one state that would have resolved this combination scenario. So it's unclear as to whether the badge color and focus color _just happen to match_ or are _intentionally the same_. Without understanding the _intent of the design_, I'm hesitant to deviate or extrapolate a conclusion.
Screen Shot 2019-11-27 at 11 32 31 AM

ryanjduffy:
I agree. We should ask that question of the design team.

Djspaceg:
> As an aside, it appears there's a styling regression in `IconButton` in electro. We need not fix that here but wanted to note it for future follow up.

Could you provide more details about this regression? And is this specific to the literal component called `IconButton` or is it also present in `<Button icon="closex" />`?

ryanjduffy:
Sorry! Forgot to include specifics. The `border-radius` set causing it to look more oval-ish rather than the parallelogram. I didn't bisect to determine where that came up.

ryanjduffy:
I'll open a new ticket to investigate it.

Djspaceg:
This is a one-line regression from a previous recent ticket.

Djspaceg:
Resolution in the following PR: https://github.com/enactjs/agate/pull/179

Djspaceg:
I'm unclear as to what the blocking change request is for this PR. Is there a specific approach you had in mind?

Regarding the fixed color assignment in a JS prop (which admittedly is fully against the dynamic and semantic nature of the intent of this repo), we don't presently have a better approach implemented. There are at least 2 closer-to ideal paths that I can see right now.
1. We implement a keyword system that allows a developer to choose one of a limited set of intent-based semantic words that a skin is able to map to specific colors internally. This retains the semantic nature and allows full freedom for the skin to choose how to represent in its own visual color language. The drawback being that an app developer must _trust_ that the skin will do what's best for itself, rather than the app developer able to impose preference from an app level.
2. We implement a library of pre-defined colors, which a skin can tune to match the specifc saturation and lightness levels that make that color visually fit in with the other colors of the skin. This approach is most like what Moonstone does for the `Button` `color` prop, where 4 color words are available and the moonstone skin defines what exactly those colors mean. The color value can be updated over the years while the developer still just specifies "red", "blue", etc. For example, Carbon may define "red" as `#ff0000` (full aggressive red) while Titanium may define "red" as `#ee6666` (softer balanced red). The palette of swatches here can be as many as we choose, but will be more visually restrictive in terms of how the skin is able to customize the values, since they'll be tied to real color concepts. E.g. "red" cannot be represented by `#00cc00`, even though that color may better fit in with the skin.

My personal preference is to implement #1 above, and do it at a soon but later time. It better fits in with the intent of Agate and provides flexibility to the visual designer responsible for the skin.

ryanjduffy:
My opinion is that, at a minimum, the sizes and positions should be updated across skins and `size` values. Historically, when we add API to a UI library we expect that it works in all supported scenarios. If we're not ready to do that for this, we might want to consider an app-side solution until  the agate-side solution can be fully built.

Djspaceg:
I'll add designs to the other skins and sizes now.

Djspaceg:
Now dependent on #177 and #179.

Djspaceg:
Custom badge sizes is now implemented.

```
import Button from '@enact/agate/Button';

import Skinnable from '@enact/agate/Skinnable';

const SkinFrame = Skinnable((props) => <div {...props}>
	<Button badge="123" size="huge">A</Button>
	<Button badge="123" size="large">A</Button>
	<Button badge="123" size="large" icon="closex" />
	<Button badge="123" size="small">B</Button>
	<Button badge="123" size="small" icon="closex" />
	<Button badge="123" size="smallest">C</Button>
	<Button badge="123" size="smallest" icon="closex" />
</div>);
```
Screen Shot 2019-11-27 at 1 30 04 PM

teckliew:
Generally looks 👍
We can revisit the color focus issue after the designs resolve that issue.

PR #177, opened 2019-11-26
PLAT-95788: Update LabeledIconButton to use Button for icon

JayCanuck:
Switches from `IconButtonBase` to `ButtonBase` as the foundation for `LabeledIconButton`.

Additionally supports the forwarding of the breadth of props for Button, to allow such features as Spinners in LabeledIconButtons, etc..

Enact-DCO-1.0-Signed-off-by: Jason Robitaille <jason.robitaille@lge.com>


LabeledIconButton/LabeledIconButton.js
```diff
@@ -76,11 +97,29 @@ const LabeledIconButtonBase = kind({
...
 			icon: (
-				<Button pressed={pressed} selected={selected} spriteCount={spriteCount}>{icon}</Button>
+				<Button
+					backgroundOpacity={backgroundOpacity}
+					icon={icon}
+					iconComponent={iconComponent}
+					highlighted={highlighted}
+					pressed={pressed}
+					selected={selected}
+					spriteCount={spriteCount}
```
ryanjduffy: Is this prop gone now? I'm seeing warnings about it and can't find references any longer.

Djspaceg: This prop needs to be restored onto Button and sent along to the inner Icon component. I think we may need to update enact/ui to properly add this.

ryanjduffy: How do you want to proceed for this PR?

Djspaceg: I will spend a few minutes and see if I can make it joyful again.

Djspaceg: Let's ignore this for now and address it later. I'm not having any luck fixing it efficiently.

PR #180, opened 2019-12-02
PLAT-95265: Remove IconButton from Agate

RenukaAtale:
- Remove IconButton files
- Update LabeledIcon and LabelIconButton to use Button instead of IconButton
- Remove all the css related to IconButton in different skins and add it to Button
- Replace IconButton with Button in all the files using IconButton


Panels/Controls.js
```diff
@@ -93,7 +93,7 @@ const ControlsBase = kind({
  * ```
  * // remove the close button and use a star icon button
  * <Controls noCloseButton>
- *   <IconButton>star</IconButton>
+ *   <Button icon="star"></Button>
```
sugardave: Since this has no `children`, just make it `<Button icon="star" />`

styles/variables-base.less
```diff
@@ -448,4 +428,4 @@
 @agate-togglebutton-indicator-border-width: 0;
 @agate-togglebutton-indicator-border-radius: @agate-button-border-radius;
 @agate-togglebutton-indicator-height: 6px;
-@agate-togglebutton-indicator-transform: none;
+@agate-togglebutton-indicator-transform: none;
```
sugardave: Let's put the newline at the end of the file back

ToggleButton/ToggleButton.js
```diff
@@ -150,18 +148,19 @@ const ToggleButtonBase = kind({
...
 		delete rest.toggleOffLabel;
 		delete rest.toggleOnLabel;
 
-		return iconToggleButton ? (
-			<IconButton {...rest} aria-pressed={selected} css={css} selected={selected}>{icon}</IconButton>
-		) : (
-			<Button {...rest} aria-pressed={selected} css={css} icon={icon} selected={selected}>{children}</Button>
+		return (
+			<Button {...rest} aria-pressed={selected} css={css} icon={icon} selected={selected}>
+				{children}
```
sugardave: I think we need to incorporate the `if (iconToggleButton)` logic in the props of this button, maybe?

Something like
```
<Button {...rest} aria-pressed={selected} css={css} icon={iconToggleButton ? icon : null} selected={selected}>
    {iconToggleButton ? null : children}
</Button>
```
Or, if we know that we want to potentially have both, remove the computed `iconToggleButton` completely.

RenukaAtale: Removed iconToggleButton logic as button already has the logic to handle the icon with children.

Scrollable/ScrollButton.js
```diff
@@ -4,17 +4,17 @@ import PropTypes from 'prop-types';
...
 import ReactDOM from 'react-dom';
 
-import IconButton from '../IconButton';
+import Button from '../Button';
 
 import css from './Scrollbar.module.less';
 
 /**
- * An [IconButton]{@link agate/IconButton.IconButton} used within
+ * An [Button]{@link agate/Button.Button} used within
```
sugardave: Minor: Change `"An [Button]..."` to `"A [Button]..."`

Djspaceg:
I thought we had planned on not deleting the `IconButton` component yet, but would retain it for a short time. We decided against that, I take it?

ryanjduffy:
I was unaware of that plan. Was it documented somewhere?

Djspaceg:
Only through conversation in person.

Djspaceg:
It's probably fine. No action needed now. It just seemed like it would be a safer path. I'll be more thorough when setting plans in the future, rather than just talking to people.

PR #181, opened 2019-12-04
ENYO-6364: Updated Button to include Icon-specific

Djspaceg:
Updated Button to include Icon-specific and allow icon prop of LabeledIconButton to be a boolean.

----

Now that Button is being used instead of IconButton, Button needs a few feature props added to bring it back in-line with IconButton's customization capabilities.

LaabeledIconButton also needs to be able to support the Boolean type for `icon` so the `Spinner` component can be passed in as the `iconComponent`. The following may be useful to anyone wanting to inject a Spinner as an Icon into a LabeledIconButton:
```
const IconableSpinner = ({...rest}) => {
	delete rest.spriteCount;
	return <Spinner {...rest} />;
};
```


Button/Button.js
```diff
@@ -196,6 +214,15 @@ const ButtonBase = kind({
...
 			);
 		},
+		iconComponent: ({iconComponent, spriteCount}) => {
+			// Don't burden basic HTML elements with the spriteCount prop (or other Icon-specific props)
+			if (typeof iconComponent === 'string') return iconComponent;
+
+			return (<ComponentOverride
+				component={iconComponent}
+				spriteCount={spriteCount}
+			/>);
```
sugardave: Format :yuck:

Djspaceg: Should be better now

sugardave: :awthanks:

PR #186, opened 2019-12-11
PLAT-87509: Add New Icons

teckliew:
Added 10 new icons and updated 1
agate-icon-update


Icon/IconList.js
```diff
@@ -22,6 +22,8 @@ export default {
 	arrowlargeright   : 0x0EFFF0, // caret_right_large
 	closex            : 0x0EFFF5, // close
 	search            : 0x0EFFF6, // search
+	bluetooth2        : 0x0F000A, // bluetooth2
```
webOS101: In the future, after docs branch is merged, please also update the icon list in the docs section of `Icon.js`.  I updated the current docs branch with these new icons.

PR #187, opened 2019-12-13
PLAT-96102: Add CheckboxItem

teckliew:
### Feature

Add `CheckboxItem`

checkboxdark
checkboxlight

### Considerations

Moved `ToggleIcon` to `internal` for use in both `Switch` and `Checkbox`


germboy:
I like this implementation overall. :nice: job
I only had a couple of concerns:
- When passing `iconPosition="after"`, the checkbox will display immediately after the text/children (as if we're using `inline`), rather than aligning to the right side of the `CheckboxItem`.
- Should there be padding/margin on the outside side of the checkbox? I didn't see GUI specs for this, so it may be a judgement call. I notice for a standard agate `Item` or `LabeledIcon` (even moonstone `CheckboxItem`), there is some checkbox padding to help better control the layout so it seems to make sense to add some here as well. This could also go along with the above concern for when using `iconPosition="after"`.
- The checkbox background/bounding box is only styled and visible for the Gallium themes. Is this intended?

CheckboxItem/CheckboxItem.js
```diff
@@ -0,0 +1,100 @@
...
+	},
+
+	styles: {
+		css: componentCss,
+		className: 'checkboxItem',
+		publicClassNames: ['checkboxItem']
+	},
+
+	computed: {
+		children: ({children}) => <Cell align="center">{children}</Cell>
```
ryanjduffy: Curious about this. There isn't really a computation here so why did you choose to add this versus including it below in render?

CheckboxItem/CheckboxItem.js
```diff
@@ -0,0 +1,100 @@
...
+	},
+
+	render: (props) => (
+		<ToggleItem
```
ryanjduffy: I'm not sure we need the abstraction of ToggleItem + SlotItem + Item after the changes in #42. `SwitchItem` uses `Item` directly so I'm curious if that's possible here as well. Can you look into that option a bit?

Here's a rough branch to give you an idea of what I was thinking.

https://github.com/enactjs/agate/compare/feature/PLAT-96102...feature/PLAT-96102-ryan?expand=1

CheckboxItem/Checkbox.js
```diff
@@ -0,0 +1,75 @@
...
+ * <Checkbox selected />
+ * ```
+ *
+ * @class Checkbox
+ * @memberof agate/Checkbox
+ * @extends agate/ToggleIcon.ToggleIcon
+ * @ui
+ * @private
+ */
+const CheckboxBase = kind({
```
ryanjduffy: Let's move this to a top-level component like `Switch`. `ToggleIcon` might be a useful private top-level or `internal` component to share between `Switch` and `Checkbox` too. Please check that out and decide if it's useful or not.

Switch/Switch.js
```diff
@@ -27,7 +23,11 @@ import componentCss from './Switch.module.less';
  *
  * @class SwitchBase
  * @memberof agate/Switch
+<<<<<<< HEAD
```
ryanjduffy: Oops :)

Item/Item.js
```diff
@@ -100,32 +100,30 @@ const ItemBase = kind({
...
-				{slotAfter ? (
-					<Cell className={css.slotAfter} shrink>
-						{slotAfter}
-					</Cell>
-				) : null}
-			</UiItemBase>
-		);
-	}
+	render: ({children, componentRef, css, label, labelPosition, slotAfter, slotBefore, ...rest}) => (
+		<UiItemBase {...rest} css={css} align="center" component={Row} ref={componentRef}>
```
ryanjduffy: Prefer we keep the `return` since we nearly always need to delete props eventually and then we'll have add it back. Using from the start minimizes the diff later which simplifies reviews and bisection.

CheckboxItem/CheckboxItem.js
```diff
@@ -0,0 +1,133 @@
...
+	},
+
+	render: ({children, css, icon, iconPosition, selected, ...rest}) => (
+		<Item
+			data-webos-voice-intent="SelectCheckItem"
+			role="checkbox"
+			{...rest}
+			css={css}
+		>
+			{iconPosition === 'after' ? (<Checkbox selected={selected} slot="slotAfter">{icon}</Checkbox>) : null}
```
ryanjduffy: We can simplify this and L116 by moving the conditional into `slot` since the order isn't relevant when using `Slottable`; it'll pluck them into the right slot regardless of order.

```jsx
<Checkbox selected={selected} slot={iconPosition === 'after' ? 'slotAfter' : 'slotBefore'}>{icon}</Checkbox>
```

CheckboxItem/CheckboxItem.js
```diff
@@ -0,0 +1,133 @@
...
+		icon: PropTypes.oneOfType([PropTypes.element, PropTypes.func, PropTypes.string]),
+
+		/**
+		 * Specifies on which side (`'before'` or `'after'`) of the text the checkbox appears.
+		 *
+		 * @type {('before'|'after')}
+		 * @default 'before'
+		 * @public
+		 */
+		iconPosition: PropTypes.string,
```
ryanjduffy: Do we have requirements for this? I think it has value but I'm concerned about the inconsistency of it only appearing here and not in the other toggle items (`RadioItem`, `SwitchItem`). If we lack specific requirements, let's defer supporting it for now so we can do so consistently.

CheckboxItem/CheckboxItem.js
```diff
@@ -0,0 +1,133 @@
...
+		css: PropTypes.object,
+
+		/**
+		 * Customize the component used as the check in the checkbox.
+		 *
+		 * @type {Element|Function|String}
+		 * @default 'check'
+		 * @public
+		 */
+		icon: PropTypes.oneOfType([PropTypes.element, PropTypes.func, PropTypes.string]),
```
ryanjduffy: I think this needs to match `agate/Icon.children`'s signature. I tracked through the source and couldn't see a way to use a component or element but please let me know if I missed it.

CheckboxItem/CheckboxItem.module.less
```diff
@@ -0,0 +1,31 @@
...
+
+		.slotBefore {
+			-webkit-margin-start: 0;
+			-webkit-margin-end: @agate-checkboxitem-icon-margin;
+		}
+
+		.focus({
+			color: @agate-checkboxitem-focus-text-color;
+			background-color: @agate-checkboxitem-focus-bg-color;
+			border-color: white;
```
ryanjduffy: Do we need to make this a variable or should it be constant?

Checkbox/Checkbox.module.less
```diff
@@ -0,0 +1,35 @@
...
+			height: @size;
+		}
+
+		&.selected {
+			.icon {
+				visibility: visible;
+			}
+		}
+
+		color: @agate-checkbox-text-color;
```
ryanjduffy: We've got some disjointed rules here. Let's move these up with the others and collapse the two `.icon` blocks.

Checkbox/Checkbox.module.less
```diff
@@ -0,0 +1,35 @@
+// Checkbox.module.less
+//
+@import "../styles/mixins.less";
+@import "../styles/skin.less";
+
+.toggleIcon {
+	.applySkins({
+		@size: @agate-checkbox-size;
+
+		width: @size;
```
ryanjduffy: This might be too restrictive for agate. I'd be a bit surprised if a skin wanted a checkbox that wasn't the same size in both directions but they could. I'm okay to leave this but wanted to note for future consideration.

ryanjduffy:
Pushed a commit to move the Checkbox LESS variables under their own header.

CheckboxItem/CheckboxItem.module.less
```diff
@@ -0,0 +1,32 @@
+// Checkbox.module.less
+//
+@import "../styles/mixins.less";
+@import "../styles/skin.less";
+
+.checkboxItem {
+	align-items: center;
+
+	.applySkins({
+		height: @agate-checkboxitem-height;
```
ryanjduffy: @teckliew Curious if this is really necessary. When I added the label knobs to the sampler for this, the fixed height causes layout issues when the label is above or below. Does `CheckboxItem` need its own height value or should it rely on `Item`?

teckliew: We should leave it up to `Item`. This was leftover from before we use `Item`. Good catch.

PR #188, opened 2019-12-13
PLAT-97799: Add Skin Variants for night mode

Djspaceg:
Converted existing skins to use `skinVariants` to capture day vs night mode. All skins can be easily switched between and their appropriate day or night skin will switch along with the main one.

A companion branch of the same name is available in the https://github.com/enyojs/webos-auto-settings repo.

### DIFF
--- a/AgateDecorator/AgateDecorator.js
+++ b/AgateDecorator/AgateDecorator.js
@@ -217,7 +217,7 @@ const AgateDecorator = hoc(defaultConfig, (config, Wrapped) => {
 	if (ri) App = ResolutionDecorator(ri, App);
 	if (i18n) App = I18nDecorator(App);
 	if (spotlight) App = SpotlightRootDecorator({noAutoFocus}, App);
-	if (skin) App = Skinnable({defaultSkin: 'gallium-day'}, App);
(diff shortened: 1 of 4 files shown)


teckliew:
Should note the addition of `Panels`'s `cover` prop that determines fullscreen or partial panel. Otherwise LGTM

AgateDecorator/AgateDecorator.js
```diff
@@ -118,7 +118,7 @@ const AgateDecorator = hoc(defaultConfig, (config, Wrapped) => {
 	if (ri) App = ResolutionDecorator(ri, App);
 	if (i18n) App = I18nDecorator(App);
 	if (spotlight) App = SpotlightRootDecorator({noAutoFocus}, App);
-	if (skin) App = Skinnable({defaultSkin: 'gallium-day'}, App);
+	if (skin) App = Skinnable({defaultSkin: 'gallium'}, App);
```
webOS101: Should this accept variants here as well?

ryanjduffy: Perhaps but only if we want to allow overriding the supported variants. `skinVariants` is an acceptable prop on the App with `Skinnable` applied so an app can also set the variants that way iirc.

PR #189, opened 2019-12-16
PLAT-95484: Update Slider

teckliew:
Updated `Slider` colors and knob position.

slider-day-focus
slider-day
slider-night-focus
slider-night

When focused, knob grows from 35px to 45px in height, and 5px over the edge of the progressbar:
knob-focus-edge
Normal knob at the edge of the progressbar:
knob-edge

### DIFF
--- a/styles/variables-gallium.less
+++ b/styles/variables-gallium.less
@@ -110,9 +110,9 @@
 @agate-slider-knob-border-radius: (@agate-slider-knob-height / 2);
 @agate-slider-knob-resting-state-scale: 0.77;
 // Additional offset (-12% at min position, 0% halfway, and +12% at max position) is needed when scaled down to 0.77 so that the knob stays within the progressbar
-@agate-slider-knob-transform: translate(calc((var(--ui-slider-proportion-end-knob) * -100%) + (-12% + ((var(--ui-slider-proportion-end-knob) * 0.24) * 100%))), -50%) scale(0.77);
+@agate-slider-knob-transform: translate(calc((var(--ui-slider-proportion-end-knob) * -100%) + (-12% + (var(--ui-slider-proportion-end-knob) * 24%))), -50%) scale(@agate-slider-knob-resting-state-scale);
 // Additional offset (-5px at min position, 0px halfway, and +5px at max position) is needed when focused and scaled back to 1
-@agate-slider-knob-active-transform: translate(calc((var(--ui-slider-proportion-end-knob) * -100%) + (-5px + ((var(--ui-slider-proportion-end-knob) * 0.1) * 100px))), -50%) scale(1);
+@agate-slider-knob-active-transform: translate(calc((var(--ui-slider-proportion-end-knob) * -100%) + (-5px + (var(--ui-slider-proportion-end-knob) * 10px))), -50%) scale(1);
 @agate-slider-border-radius: 3px;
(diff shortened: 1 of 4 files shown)


Djspaceg:
Is the growing on focused/pressed state necessary? As in, is it part of the design? I didn't remember seeing that.

teckliew:
Yea it’s part of the design that focused slider knob is 10px bigger; and if the knob is at the edge of the progress bar, the knob is 5px over the edge when focused.

Djspaceg:
Oh boy! That's cumbersome haha. Glad you noticed that!

styles/variables-gallium.less
```diff
@@ -106,10 +106,14 @@
...
 // Slider
 // ---------------------------------------
+@agate-slider-knob-height: 45px;
 @agate-slider-knob-border-radius: (@agate-slider-knob-height / 2);
-@agate-slider-knob-resting-state-scale: 0.8;
-@agate-slider-knob-transform: @agate-translate-center scale(0.8);
-@agate-slider-border-radius: 0;
+@agate-slider-knob-resting-state-scale: 0.77;
+// Additional offset (-12% at min position, 0% halfway, and +12% at max position) is needed when scaled down to 0.77 so that the knob stays within the progressbar
+@agate-slider-knob-transform: translate(calc((var(--ui-slider-proportion-end-knob) * -100%) + (-12% + ((var(--ui-slider-proportion-end-knob) * 0.24) * 100%))), -50%) scale(0.77);
```
Djspaceg: Let's use `@agate-slider-knob-resting-state-scale` in the `scale()` section instead of that `0.77` literal. It doesn't seem to be referenced by any file outside of the Variables files.

Also, I'm intrigued by the `0.24` here and the `0.1` down below. Is the 24% used to get the -12% and +12%? If so, just for readability, let's change it from `0.24` to `24%` the math should still all work since the units are still in %.
As for the `0.1` down below, that seems more complex as it relates to a percentage rather than the -5/+5 px necessary to keep it positioned well. Could you try this out on sliders of different widths to verify they look as expected? If they don't perhaps switching the fixed 5px to an equivalent % would work too?

teckliew: > As for the 0.1 down below, that seems more complex as it relates to a percentage rather than the -5/+5 px necessary to keep it positioned well. Could you try this out on sliders of different widths to verify they look as expected? If they don't perhaps switching the fixed 5px to an equivalent % would work too?

The math works with different widths because the offset is constant to the scale of the knob, and the offset is also is a constant +-5px when focused.

styles/variables-gallium.less
```diff
@@ -106,10 +106,14 @@
 
 // Slider
 // ---------------------------------------
+@agate-slider-knob-height: 45px;
 @agate-slider-knob-border-radius: (@agate-slider-knob-height / 2);
-@agate-slider-knob-resting-state-scale: 0.8;
-@agate-slider-knob-transform: @agate-translate-center scale(0.8);
-@agate-slider-border-radius: 0;
+@agate-slider-knob-resting-state-scale: 0.77;
+// Additional offset (-12% at min position, 0% halfway, and +12% at max position) is needed when scaled down to 0.77 so that the knob stays within the progressbar
```
ryanjduffy: Thanks for the inline docs here, @teckliew !

PR #191, opened 2020-01-04
ENYO-6393: Fix up SliderButton

Djspaceg:
SliderButton was largely ignored after its initial creation. This PR aims to improve its customizability by a skin, and leverages that customizability to make the SliderButton component, when used in the Gallium skin, look like it matches the target visual style.

SliderButton-BeforeAfter

### DIFF
--- a/SliderButton/SliderButton.js
+++ b/SliderButton/SliderButton.js
@@ -45,6 +45,7 @@ const SliderKnob = kind({
 	},
 	render: ({...rest}) => {
 		delete rest.value;
+		delete rest.tooltipComponent;
 		return (
 			<div
 				{...rest}
@@ -57,14 +58,15 @@ const SliderProgress = kind({
 	name: 'SliderProgress',
 	propTypes: {
 		css: PropTypes.object,
-		values: PropTypes.arrayOf(PropTypes.number)
+		values: PropTypes.arrayOf(PropTypes.string)
 	},
 	styles: {
 		css: componentCss,
 		className: 'track'
 	},
 	render: ({children, css, values, ...rest}) => {
-
+		delete rest.progressAnchor;
+		delete rest.backgroundProgress;
 		return (
 			<Row {...rest} align="center">
 				{children}
--- a/SliderButton/SliderButton.js
+++ b/SliderButton/SliderButton.js
@@ -45,6 +45,7 @@ const SliderKnob = kind({
 	},
 	render: ({...rest}) => {
 		delete rest.value;
+		// eslint-disable-next-line enact/prop-types
 		delete rest.tooltipComponent;
 		return (
 			<div
@@ -65,7 +66,9 @@ const SliderProgress = kind({
 		className: 'track'
 	},
 	render: ({children, css, values, ...rest}) => {
+		// eslint-disable-next-line enact/prop-types
 		delete rest.progressAnchor;
+		// eslint-disable-next-line enact/prop-types
 		delete rest.backgroundProgress;
 		return (
 			<Row {...rest} align="center">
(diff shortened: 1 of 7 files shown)


styles/colors-gallium-day.less
```diff
@@ -215,14 +215,16 @@
 
 // SliderButton
 // ---------------------------------------
-@agate-sliderbutton-color:          @agate-accent;
+@agate-sliderbutton-color:          @agate-white;
```
teckliew: Maybe a darker `@agate-sliderbutton-color` would make the text more legible for Gallium Day? Right now it's a bit hard to read white text on light gray.

SliderButton

SliderButton/SliderButton.js
```diff
@@ -57,14 +58,15 @@ const SliderProgress = kind({
...
+		values: PropTypes.arrayOf(PropTypes.string)
 	},
 	styles: {
 		css: componentCss,
 		className: 'track'
 	},
 	render: ({children, css, values, ...rest}) => {
-
+		delete rest.progressAnchor;
+		delete rest.backgroundProgress;
```
teckliew: Looks like we need to clean up `ui/Slider` to avoid unnecessary deleting props both here and the slider knob.

Djspaceg: I think the API "contract" established in `ui/Slider` is appropriate. Tooltips and `backgroundProgress` etc do make sense to forward into these specialized components (in any theme). This component in particular, `SliderButton` is leveraging several of the qualities of `ui/Slider` but it is also providing its own components for the building blocks (internal components `SlideKnob` and `SliderProgress`). Each of those building blocks needs to respect its role and handle the incoming props accordingly, so it doesn't warn or break. This seems fine to me.

I've added linting ignore lines for these props so we don't have to name them in the proptypes. _It always seems weird to declare a prop, just to delete it later without doing anything with or to it._

styles/colors-gallium-day.less
Djspaceg: I updated these to be `@agate-text-color` in both skins, which seems more appropriate. Declaring plain white, like I did before, seemed wrong, while this seems more in-line with the skins' core approach.

styles/variables-gallium.less
```diff
@@ -150,6 +150,12 @@
 
 // SliderButton
 // ---------------------------------------
+@agate-sliderbutton-border-radius: 3px;
+@agate-sliderbutton-border-width: 12px;
+@agate-sliderbutton-bg-clip: padding-box;
+@agate-sliderbutton-knob-border-radius: 6px;
+@agate-sliderbutton-knob-border-width: 3px;
+@agate-sliderbutton-knob-transform: scale(1.07, 1.4);
```
teckliew: What do you think about having a smaller scale like `scale(1.04, 1.4)`, and also add a top margin as well? So that the knob shadow won't be clipped.
sliderbuttonclip

SliderButton/SliderButton.js
ryanjduffy: We might consider declaring these as prop types but marking them `@private` with docs that indicate why they exist. That'll give us better traceability for the rationale (which i think is sound) than an otherwise undocumented `eslint-disable` comment.

Djspaceg: If at very least, the docs for these could read:
> Unused, deleted in render.

I guess that's good practice; it just seems cumbersome.

ryanjduffy: I'm most concerned with clarity for the reader. I think that describes _what_ and not _why_. The _what_ is relatively clear from the code but the _why_ isn't. I think I'd favor some text that clarifies that these props are required by the `ui/Slider` API but are not required for this implementation.

Djspaceg: I follow now. That makes sense.

PR #193, opened 2020-01-09
PLAT-95484: Additional Slider knob offset fix

teckliew:
Adding onto https://github.com/enactjs/agate/pull/189
Needed to move the knob offset to the knob's container instead of translating the knob itself, so that the knob won't animate the position offset.

Readjusted the `ProgressBar`'s width and margin so that `PositionDecorator` will calculate correctly given the knob's offset.

The math is: the knob needs to be +18px at min and -18px at max to be inside the progress bar. Added additional 18px to the progress bar on both sides, and then had to resort to negative 18px margin to center the progress bar.

### DIFF
--- a/Slider/Slider.module.less
+++ b/Slider/Slider.module.less
@@ -83,7 +83,8 @@
 			}
 
 			.knob {
-				left: ~"calc(var(--ui-slider-proportion-end-knob) * 100%)";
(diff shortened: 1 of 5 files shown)


ryanjduffy:
Closing in favor of #195

PR #194, opened 2020-01-14
PLAT-98187 [Auto Agate] Add Dropdown component

RenukaAtale:
Added dropdown component to Agate


Dropdown/Dropdown.js
```diff
@@ -0,0 +1,195 @@
...
+	},
+
+	styles: {
+		css: componentCss,
+		className: 'Dropdown',
+		publicClassNames: true
+	},
+
+	computed: {
+		className: ({css, selected, styler}) => styler.append(selected && css.selected),
```
teckliew: I don't think we have a `selected` style. Although maybe we should have a selected style for the selected option in the dropdown list.

Dropdown/Dropdown.js
```diff
@@ -0,0 +1,195 @@
...
+	},
+
+	defaultProps: {
+		direction: 'down',
+		open: false
+	},
+
+	styles: {
+		css: componentCss,
+		className: 'Dropdown',
```
teckliew: Lowercase for css class name: `dropdown`

Dropdown/Dropdown.js
```diff
@@ -0,0 +1,195 @@
...
+		 *
+		 * @type {Function}
+		 */
+		onSelect: PropTypes.func,
+		/*
+		 * Called when clicked on the dropdown to open.
+		 *
+		 * @type {Number}
+		 */
+
```
teckliew: Spacing. Also there should be a line between each prop type block.

Dropdown/Dropdown.js
```diff
@@ -0,0 +1,195 @@
...
+		/**
+		 * Customizes the component by mapping the supplied collection of CSS class names to the
+		 * corresponding internal elements and states of this component.
+		 *
+		 * @type {Object}
+		 * @public
+		 */
+		css: PropTypes.object,
+		/**
+		 * The animation direction of the `palette`.
```
teckliew: Maybe update the doc to "the direction where the dropdown list appears"?

Dropdown/Dropdown.js
```diff
@@ -0,0 +1,195 @@
...
+
+	styles: {
+		css: componentCss,
+		className: 'Dropdown',
+		publicClassNames: true
+	},
+
+	computed: {
+		className: ({css, selected, styler}) => styler.append(selected && css.selected),
+		transitionContainerClassname: ({css, open, styler}) => styler.join(css.transitionContainer, (open ? css.openTransitionContainer : null)),
```
teckliew: We could clean this up to just be `styler.join(css.transitionContainer, {css.openTransitionContainer : open})`

Dropdown/Dropdown.js
```diff
@@ -0,0 +1,195 @@
...
+				</Transition>
+			</div>
+		);
+	}
+});
+
+
+
+/**
+ * Applies Agate specific behaviors to [ColorPickerBase]{@link agate/Dropdown.DropdownBase}.
```
teckliew: `ColorPickerBase` should be updated to `DropdownBase`

Dropdown/Dropdown.module.less
```diff
@@ -0,0 +1,39 @@
+// Dropdown.module.less
+//
+@import "../styles/mixins.less";
+@import "../styles/skin.less";
+
+.Dropdown {
+	.applySkins({
+		.item {
```
teckliew: Do we need to override the item styles or should we just keep the default item styles?

Dropdown/Dropdown.module.less
```diff
@@ -0,0 +1,39 @@
...
+			border-color: gray;
+			border-style: solid;
+			border-width: 1px;
+		}
+		.transitionContainer {
+			position: absolute;
+			display: inline-block;
+			z-index: 1;
+			pointer-events: none;
+			min-width: 312px;
```
teckliew: We should probably have this as a less variable, along with all the other properties like the background color and border styles in `.dropdownList`, and the height in `.dropdownListWithScroller`

Dropdown/Dropdown.module.less
```diff
@@ -0,0 +1,39 @@
...
+			width: 100%;
```
teckliew: Should we keep this to 100%? Or how should we support different dropdown widths?

RenukaAtale: At this point we need keep the width,  don't have enough information for different sizes for dropdown.

Dropdown/Dropdown.module.less
RenukaAtale: I have added the style on the first item for border. we won't be able to use the default item style as don't have default border on the item.

Dropdown/Dropdown.js
```diff
@@ -0,0 +1,200 @@
...
+	computed: {
+		transitionContainerClassname: ({css, open, direction, styler}) => styler.join(css.transitionContainer, {openTransitionContainer: open, upTransitionContainer: direction === 'up'} ),
+		dropdownListClassname: ({children, css, styler}) => styler.join(css.dropdownList, {dropdownListWithScroller: children.length > 4}),
+		title: ({children, selected}) => {
+			if (isSelectedValid({children, selected})) {
+				const child = children[selected];
+				return typeof child === 'object' ? child.children : child;
+			}
+
+			return children[0];
```
webOS101: `children` is not required.  Will this throw if `children` is `null`?

Dropdown/Dropdown.js
```diff
@@ -0,0 +1,200 @@
...
+		 */
+		css: PropTypes.object,
+
+		/**
+		 * The direction where the dropdown list appears.
+		 *
+		 * @type {String}
+		 * @public
+		 */
+		direction: PropTypes.string,
```
webOS101: Can you enumerate the valid values here and in the jsdoc?

Dropdown/Dropdown.js
```diff
@@ -0,0 +1,200 @@
...
+ * @memberof agate/Dropdown
+ * @ui
+ * @public
+ */
+const DropdownBase = kind({
+	name: 'Dropdown',
+
+	propTypes: /** @lends agate/Dropdown.Dropdown.prototype */ {
+		/*
+		 * The selections for Dropdown
```
webOS101: Should we mention that the title will be the first item if no items are selected (which seems odd to me, is that per the design)?

Dropdown/Dropdown.js
```diff
@@ -0,0 +1,200 @@
...
+
+		/*
+		 * Called when an item is selected.
+		 *
+		 * @type {Function}
+		 */
+		onSelect: PropTypes.func,
+
+		/*
+		 * Called when clicked on the dropdown to open.
```
webOS101: This is not a callback, but the open state.  Add the `@default` value here?

Dropdown/Dropdown.js
```diff
@@ -0,0 +1,200 @@
...
+		transitionDirection: ({direction}) => {
+			switch (direction) {
+				case 'left':
+					return 'right';
+				case 'right':
+					return 'left';
+				case 'up':
+					return 'down';
+				case 'down':
+					return 'up';
```
webOS101: Should we add `default:` here?

Dropdown/Dropdown.js
```diff
@@ -0,0 +1,200 @@
...
+							>{children || []}</Group>
+						</Scroller>
+					</ContainerDiv>
+				</Transition>
+			</div>
+		);
+	}
+});
+
+
```
webOS101: Can delete these extra empty line here and below.

Dropdown/Dropdown.js
```diff
@@ -0,0 +1,200 @@
...
+ *
+ * By default, `Dropdown` maintains the state of its `selected` property. Supply the
+ * `defaultSelected` property to control its initial value. If you wish to directly control updates
+ * to the component, supply a value to `selected` at creation time and update it in response to
+ * `onSelected` events.
+ *
+ * @class Dropdown
+ * @memberof agate/Dropdown
+ * @extends agate/Dropdown.DropdownBase
+ * @ui
```
webOS101: `@mixes agate/Dropdown.DropdownDecorator`

Dropdown/Dropdown.js
```diff
@@ -0,0 +1,200 @@
...
+		 *
+		 * @type {String[]|Array.<{key: (Number|String), children: (String|Component)}>}
+		 */
+		children: PropTypes.array,
+
+		/**
+		 * Customizes the component by mapping the supplied collection of CSS class names to the
+		 * corresponding internal elements and states of this component.
+		 *
+		 * @type {Object}
```
webOS101: Are there any customizable classes?  If not, maybe we should just mark this private?  If there are, can you document them here?

Dropdown/Dropdown.js
```diff
@@ -0,0 +1,200 @@
...
+ *
+ * @class DropdownBase
+ * @memberof agate/Dropdown
+ * @ui
+ * @public
+ */
+const DropdownBase = kind({
+	name: 'Dropdown',
+
+	propTypes: /** @lends agate/Dropdown.Dropdown.prototype */ {
```
webOS101: `DropdownBase.prototype`

Dropdown/Dropdown.js
webOS101: I think this is the opposite of what we would want, isn't it?  The transition is opposite from the opening so it should be the same as the default.

Dropdown/Dropdown.js
```diff
@@ -0,0 +1,224 @@
...
+ * @class DropdownBase
+ * @memberof agate/Dropdown
+ * @ui
+ * @public
+ */
+const DropdownBase = kind({
+	name: 'Dropdown',
+
+	propTypes: /** @lends agate/Dropdown.DropdownBase.prototype */ {
+		/*
```
webOS101: jsdoc comments need to start with `/**` or else they won't be parsed.  I have fixed these.

Dropdown/Dropdown.js
```diff
@@ -0,0 +1,222 @@
+/**
+ * Agate styled Header.
```
webOS101: oops.

PR #195, opened 2020-01-17
PLAT-98491: Gallium slider knob fully encapsulated

Djspaceg:
The Gallium slider design shows the knob always fully within the bounds of the slider track. This update extends the visible track to protrude past the measurement boundaries of the slider track, so the knob never appears to leave the track area.

This is achieved by relocating and decoupling the visuals from the "math container" of the slider. The benefit of which is that now the knob always remains centered under the drag point, and behaves as expected when the extreme edges are clicked, as well as remains stationary on repeated tapping on the bar.

One minor visual regression arose with the Electro skin, which is also corrected in this PR. Conveniently, this made the Electro configuration simpler while creating no visual differences. All other skins were unaffected by this change.

The following code can be used to add all skinned sliders to a sample app (moot, however, now that we have "all skins" knob in storybook)
```
	<Slider skin="gallium" />
	<Slider skin="carbon" />
	<Slider skin="cobalt" />
	<Slider skin="copper" />
	<Slider skin="electro" />
	<Slider skin="titanium" />
	<Slider orientation="vertical" skin="gallium" />
```

### DIFF
--- a/IncrementSlider/IncrementSlider.module.less
+++ b/IncrementSlider/IncrementSlider.module.less
@@ -31,7 +31,10 @@
 			.slider {
 				flex-grow: 1;
 				margin: @agate-incrementslider-horizontal-slider-margin;
-				border-radius: @agate-incrementslider-border-radius;
+
+				&::before {
+					border-radius: @agate-incrementslider-border-radius;
+				}
 			}
 		}
 
@@ -55,7 +58,10 @@
 				order: 2;
 				margin: @agate-incrementslider-vertical-slider-margin;
 				padding: @agate-incrementslider-vertical-slider-padding;
-				border-radius: @agate-incrementslider-border-radius;
+
+				&::before {
+					border-radius: @agate-incrementslider-border-radius;
+				}
 			}
 
 			.incrementButton {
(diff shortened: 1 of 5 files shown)


Djspaceg:
The knob wasn't meeting the end of the bar for two super tiny reasons. 1. The knob anchor size was set to the 100% size of the knob. The resting state of the knob is 90% of that, so the part of the knob you could see was 5% too small on the sides, causing a small gap. With that gap closed, the other issue, 2. was that due to rounded corners, anti-aliasing, and retina displays, a collection of sub-pixel pixels (subpixels in terms of HTML sizing, but individual pixels on the display) some remained visible. Adding 1px to the dimensions of the anchor element overcomes this issue.

As for Electro's slider, this was a bit more tricky. The proportion position of the slider is based on the width of the node. Normally this works out great, but because it's using getBoundingClientRect, which factors in transforms for measurements, an X skew will make the component measure wider. This had the effect of the bar thinking it was longer than it is _supposed to_ be, but all of the proportions still worked out. Ultimately the solution was to un-skew the parent/root node of Slider and instead skew a `::before` pseudo-element so that the visuals were identical, but the measurements could be based on the (now invisible) root element. This has the added side-effect benefit of the tappable and interactable area of the slider being a bit more intuitive and forgiving.

PR #200, opened 2020-01-30
PLAT-99203 Showing Scroll buttons in Dropdown

MikyungKim:
Replaced `ui/Scroller` to `../Scroller`
Fixed unexpected closing by clicking on scrollbar


Dropdown/Dropdown.js
```diff
@@ -121,6 +138,17 @@ const DropdownBase = kind({
 		open: false
 	},
 
+	handlers: {
```
seunghoh: We need to consider `onKeyDown`handler to close when press left or right button to close dropdown?
I am leaving the comment for future discussion

seunghoh:
LGTM. Let's discuss spotlight support in a separate ticket

PR #203, opened 2020-02-01
PLAT-99243: Add screenshot tests

ryanjduffy:
* Adds `test-ss`, `test-ss-docker`, and `test-ss-tv` to npm scripts
* Adds `tests/screenshot` folder with tests migrated from screenshot repo
* Updates `tests/ui` to reflect updates in enactjs/ui-dev-utils#77 (required to test this)


tests/screenshot/specs/AgateCopperDay-specs.js
```diff
@@ -0,0 +1,8 @@
+const {runTest} = require('@enact/ui-test-utils/utils');
+const Page = require('./AgatePage');
+
+runTest({
+	testName: 'Agate Copper Day',
+	Page: Page,
+	skin: 'copper-day'
```
webOS101: I think we'll need to update the spec runner here to pass variants down.  It may also need some additional changes in `ui-test-utils` to pass the variants down appropriately.  It's possible we can do this from the apps here, but I'll have to dig in first.

PR #215, opened 2020-02-26
PLAT-101272: Add a new prop `scrollMode`

0x64:
The new prop scrollMode is added.
- 'translate': the scroll content is scrolled by CSS translate.
- 'native': the scroll content is scrolled by scrolltop (native way).


ybsung:
LGTM

PR #220, opened 2020-03-06
PLAT-101582: Replace querySelector with refs in VirtualList and Scroller

ybsung:
### Checklist

The `querySelector` are used a lot in the VirtualList and the Scroller to access item DOM elements easily. We might avoid using DOM APIs directly.

### Resolution

1. querySelector for the DOM element with spotlight id

Instead of calling `document.querySelector(`[data-spotlight-id="${spotlightId}"]`)`, the defined ref - `scrollContainerRef` is used.

2. querySelector for the item DOM elements

In https://github.com/enactjs/enact/pull/2713, the item refs are cached in the the VirtualList's prop - `itemRefs`. So the `itemRefs` is defined with useRef([]) in the `useScroll` and the `itemRefs` is used when needed to access the item DOM elements.

3. hasFocus

The code related with the querySelector was dead code. So I just removed it.

### Additional Considerations

I found that the `numOfItems` was extracted from not `scrollContentHandle.current.state` but `scrollContentHandle.current`. So I fixed it.

```
{state: {numOfItems}, primary} = scrollContentHandle.current,
```

### Links

PLAT-101582

### Comments


MikyungKim:
Please revert the travis change.

PR #221, opened 2020-03-09
PLAT-100713: Rename internal variables

0x64:
Along with the change of internal variable names for better maintenance of `UI` and other themed libraries, updated internal variable names.
Related: PLAT-100713


ybsung:
LGTM

PR #222, opened 2020-03-10
PLAT-101422: Sync naming in VirtualList and Scroller between sandstone and agate

Baekwoo-s:
Modify the scroller and virtuallist variables in sandstone and agate to be the same.


VirtualList/useThemeVirtualList.js
```diff
@@ -304,6 +304,8 @@ const useSpottable = (props, instances, context) => {
 const useThemeVirtualList = (props) => {
 	const {itemRefs, scrollMode, scrollContainerRef, scrollContentHandle, scrollContentRef} = props;
 
+	// Mutable value
```
ybsung: We could remove the `// Mutable value`.

ybsung:
LGTM

PR #229, opened 2020-03-23
PLAT-101754: Remove the experimental feature `backdrop-filter`

Djspaceg:
`backdrop-filter` was added to `Button` as a technology demonstration and is not necessary for production at this time.


seunghoh:
👍

PR #232, opened 2020-04-02
PLAT-102575: Clean up code related scrollInfo

MikyungKim:
Cleaned up code related scrollInfo that was removed from ENYO-5710

### DIFF
--- a/Scroller/useThemeScroller.js
+++ b/Scroller/useThemeScroller.js
@@ -77,12 +77,7 @@ const useSpottable = (props, instances) => {
 	/**
 	 * Calculates the new `scrollTop`.
 	 *
-	 * @param {Node} focusedItem node
-	 * @param {Number} itemTop of the focusedItem / focusedContainer
-	 * @param {Number} itemHeight of focusedItem / focusedContainer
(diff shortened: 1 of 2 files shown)


Scroller/useThemeScroller.js
```diff
@@ -202,8 +197,6 @@ const useSpottable = (props, instances) => {
 	 * Calculates the new top and left position for scroller based on focusedItem.
 	 *
 	 * @param {Node} item node
-	 * @param {Object} scrollInfo position info. `calculateScrollTop` uses
-	 * `scrollInfo.previousScrollHeight` and `scrollInfo.scrollTop`
```
ybsung, line 206: It seems to me that there is only one param which has `item` and `scrollPosition`. But you defined two params. One for the item node. One for the scroll position.

MikyungKim: I left the `scrollPosition` because it is the parameter for this function.
`function calculatePositionOnFocus ({item, scrollPosition}) {`

ybsung: OK.

PR #234, opened 2020-04-02
PLAT-99658: Removed unused code

0x64:
There were several code fragments for overscrolling which is not supported in Agate. We don't have any requirement for it, removed unused code.


MikyungKim:
LGTM!

jeonghee27:
I get a console error when I try the wheel.

```useScroll.js:639 Uncaught TypeError: Cannot read property 'wheel' of undefined
    at HTMLDivElement.onWheel (useScroll.js:639)``` 

I used https://github.com/enactjs/enact/pull/2738/files for enact.
Did I do something wrong?

PR #235, opened 2020-04-03
PLAT-104238: Support accessibility for Picker components

Baekwoo-s:
Support accessibility for Picker components


Picker/Picker.js
```diff
@@ -63,6 +65,53 @@ const PickerBase = kind({
...
 
+		/**
+		 * Prevents read out of both the slider and the increment and decrement
+		 * buttons.
+		 *
+		 * @type {Boolean}
+		 * @memberof agate/Picker.PickerBase.prototype
+		 * @public
+		 */
+		'aria-hidden': PropTypes.bool,
```
seunghoh, line 76: Please check if we really need to define as a prop. This is the same as dom attribute and react can pass to the dom

Picker/Picker.js
```diff
@@ -116,12 +176,20 @@ const PickerBase = kind({
 						{isFirst ? '' : values[value - 1]}
 					</div>
 				</PickerButtonItem>
-				<div className={activeClassName}>
+				<div
+					aria-valuetext={typeof valueText !== 'undefined' ? valueText : values[value]}
+					className={activeClassName}
+					id={id}
```
ryanjduffy, line 182: Let's keep `id` on the root and have a derived id here (e.g. `${id}_value`) that is used by `aria-controls`. You might also want to create it in computed and only deconstruct it in render so the id lands on the root automatically.

Baekwoo-s: I fixed the id to be modified in computed.
Please tell me if this code is different from your intention.

ryanjduffy: Slightly, yes. I meant something like:

```
computed: {
  valueId: ({id}) => `${id}_value`
}
```

And then only deconstruct `valueId` from `props` so `id` lands on the root node and you use `valueId` for `aria-controls` and this node's id.

Baekwoo-s: I changed the name of the id prop according to your opinion.
Thank you.

PR #237, opened 2020-04-03
PLAT-103496: Support localization for Agate components

Baekwoo-s:
Support localization for Agate components


Spinner/Spinner.js
```diff
@@ -21,6 +21,7 @@ import UiSpinnerBase from '@enact/ui/Spinner';
 import Skinnable from '../Skinnable';
 
 import componentCss from './Spinner.module.less';
+import $L from '../../moonstone/internal/$L/$L';
```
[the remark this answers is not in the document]
baekwoo: moonstone is wrong. I modified it to use agate/internal/$L.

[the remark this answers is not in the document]
Baekwoo-s: moonstone is wrong. I modified it to use agate/internal/$L.

seunghoh:
Let's change DatePicker with iLib formatter first

seunghoh:
LGTM.
For recording our discussion, we will implement DatePicker with iLib format in the other ticket

Spinner/Spinner.js
```diff
@@ -55,7 +56,7 @@ const SpinnerCore = kind({
 	},
 
 	defaultProps: {
-		type: 'searching'
+		type: $L('searching')
```
0x64, line 59: This looks weird. This is a _public_ prop and the value should be a plane text. This value is used just to test like `type === 'searching'.

DateTimePicker/DateTimePicker.js
```diff
@@ -67,7 +59,7 @@ class DateTimePickerBase extends React.Component {
 		this.state = {
 			hour: '12',
 			minute: '00',
-			meridiem: 'AM',
+			meridiem: $L('AM'),
```
0x64, line 62: This change also affects the value to be delivered to the app through `onChange`, `onSave`, and `onClose` callback props. Is this intended?

Baekwoo-s: I agree with your opinion. Fixed to compare type with plane text.

Spinner/Spinner.js
Baekwoo-s: I agree with your opinion. Fixed to compare type with plane text.

DateTimePicker/DateTimePicker.js
Baekwoo-s: In this code, there is a problem that ilib date is not passed to App developer.
Any changes to this code will be revised in PLAT-104574.

PR #238, opened 2020-04-08
PLAT-102572: Refactored Scrollbar

MikyungKim:
Need to apply the ScrollbarBase refactoring change from ui side
ScrollThumb is a somewhat ambiguous name for the component.

### Resolution
Applied the change from ui side
Renamed `ScrollThumb` to `ScrollbarTrack`
Renamed thumb related name to scrollbarTrack.

### Additional Considerations
`ScrollbarButtons` component needs to remove render prop. We will revisit later.

### Links
PLAT-102572


useScroll/Scrollbar.js
```diff
@@ -1,112 +1,131 @@
...
+	};
+};
+
 /**
- * An Agate-styled scroller base component.
+ * An Agate-styled scrollbar base component.
  *
  * @class ScrollbarBase
  * @memberof agate/useScroll
  * @extends ui/ScrollbarBase
```
ybsung, line 49: Do we need this extends description?

MikyungKim: I think we can remove this.

useScroll/Scrollbar.js
```diff
@@ -1,112 +1,131 @@
...
-								/>
-							);
-						}}
-					/>
-				);
-			}}
-		/>
+		<div {...scrollbarProps}>
+			<ScrollButtons
+				{...restProps}
```
ybsung, line 80: I don't think that it is the good way to pass rest props to the inner React elements. Might be better to pass it the outmost div. Could you check if it is possible to do it?

useScroll/Scrollbar.js
```diff
@@ -1,112 +1,131 @@
...
+				{...scrollbarButtonsProps}
+				ref={scrollButtonsRef}
+				scrollbarTrackRenderer={() => { // eslint-disable-line react/jsx-no-bind
+					return (
+						<ScrollbarTrack
+							{...scrollbarTrackProps}
+							key="scrollbarTrack"
```
ybsung, line 87: I might be better to add the comment why we need the key property here as you were curious before.

MikyungKim: I think it's okay

useScroll/Scrollbar.js
MikyungKim: We could destruct props only for ScrollButtons I guess..

useScroll/Scrollbar.js
```diff
@@ -148,19 +160,7 @@ ScrollbarBase.defaultProps = {
...
-		'showThumb',
-		'startHidingThumb',
-		'update'
-	]}, Skinnable(ScrollbarBase)
-);
-*/
-const Scrollbar = ScrollbarBase;
+const Scrollbar = Skinnable(ScrollbarBase);
 
 Scrollbar.displayName = 'Scrollbar';
```
ybsung, line 166: We could export the `ScrolbarBase` itself instead `Scrollbar as ScrollbarBase`.

```
export {
	Scrollbar,
	ScrollbarBase
};
```

useScroll/Scrollbar.js
webOS101: :boom:

useScroll/Scrollbar.js
MikyungKim: good catch

ybsung:
LGTM

useScroll/useScrollbar.js
```diff
@@ -4,11 +4,11 @@ import {constants} from '@enact/ui/useScroll';
 const {paginationPageMultiplier} = constants;
 
 const useScrollbar = (props, instances, context) => {
```
jeonghee27, line 6: One concern is that Sandstone doesn't have a useScrollbar.js, this feature moved into useScroll.js. I understand that Agate useScrollbar has more feature because it has a PageControll button. I hope to revisit to check we don't need to unify the structure of Sandstone and Agate.

PR #244, opened 2020-04-23
PLAT-104269: Rename AgateDecorator to ThemeDecorator

sugardave:
This patch makes `AgateDecorator`'s days numbered.


webOS101:
Will need to update docs to build with this new import!

ryanjduffy:
@sugardave are we leaving AgateDecorator.module.less and screenTypes.json in AgateDecorator intentionally?

PR #249, opened 2020-05-05
PLAT-106356: ColorPicker throws a error when changing slider value

sugardave:
This patch addresses the immediate issue by making `ColorPicker` be `Changeable` so that its `value` can be managed without the HOC for the palette extension doing it.  By having a single handler (go go gadget `@enact/core/handle`!) for the base component and the HOC the different event data structures between `onChange`'s default prop of `value` and the `onSelect` of `Group` sending the value in the `data` member (which cannot be configured, feature request planned) can be reconciled.  The mismatch of props between what the HOC slider handlers were sending and the preset color group selection were sending was the root cause of this.

This patch spawned other work that I am currently doing in the [changeable-color-picker](https://github.com/enactjs/agate/tree/feature/changeable-color-picker) branch.


samples/sampler/stories/default/ColorPicker.js
```diff
@@ -17,8 +17,8 @@ storiesOf('Agate', module)
 		() => (
 			<ColorPicker
 				direction={select('direction', prop.direction, ColorPicker, 'right')}
-				defaultValue={text('defaultValue', ColorPicker, '#3467af')}
 				onChange={action('onChange')}
+				defaultValue={text('value', ColorPicker, '#3467af')}
```
ryanjduffy, line 21: Suggest we remove this knob altogether. Setting the `defaultValue` doesn't add much value for exploring how the component looks and behaves which is the purpose of the sample. We can document that the `defaultValue` may be set in the inline docs.

ColorPicker/ColorPicker.js
```diff
@@ -240,8 +238,20 @@ const ColorPickerBase = kind({
 	}
 });
 
+// This handler is meant to accommodate using `ColorPicker`'s `onChange` prop from `Changeable`
+// as the `onSelect` handler for its `Group` component that lists the set of pre-defined color
+// choices.  `onSelect` sends the chosen value in the `data` prop of its event but `Changeable`
+// expects `value`.  The slider handlers for hue, saturation, and light values in `ColorPickerExtended`
+// use this handler to update the `value` prop via `onChange` as well.
+const colorChangeHandler = handle(
```
ryanjduffy, line 246: Let's move this into the class below. It took me a while to figure out where `this.onChange` came from because it's the only handler defined outside of the class.

sugardave:
> I think our event APIs are a bit wonky on this component.

Just a little!  I haven't been able to pick back up the other branch where I was doing everything I could to get rid of the extended palette hoc to further simplify things.  It's pretty much down to only needing it for detecting clicks outside of the base component, and I've already seen a hook implementation of that somewhere, I just need to sit down and finish executing my plan for `ColorPicker`

PR #253, opened 2020-05-12
FORTECH-26-27-28-30: Added silicon skin styles for Button, Input, Radioitem, Scroller.

daniel-stoian-lgp:
**Feature Added**
Added silicon skin styles for Button, Input, Radioitem, Scroller.

**Comments**
I hope it's ok to add in this PR styles for multiple components. This won't happen in the future
This pull request is related to FORTECH-26, FORTECH-27, FORTECH-28, FORTECH-30


seunghoh:
Please don't make PR with several Jira tickets next time.

Button/Button.module.less
```diff
@@ -305,4 +306,29 @@
 			}
 		});
 	}
+
+	&:global(.spottable):active,
+	:global(.spottable):active & {
+		.applySkins({
+			background-color: transparent;
```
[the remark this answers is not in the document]
daniel-stoian-lgp: also moved the same attributes, for :focus

Button/tests/Button-specs.js
```diff
@@ -0,0 +1,51 @@
...
+			);
+
+			subject.simulate('click');
+
+			const expected = 0;
+			const actual = handleClick.mock.calls.length;
+
+			expect(actual).toBe(expected);
+		});
+
```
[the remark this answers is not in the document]
daniel-stoian-lgp: ok

Button/tests/Button-specs.js
```diff
@@ -0,0 +1,51 @@
...
+	});
+
```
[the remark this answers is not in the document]
daniel-stoian-lgp: ok

styles/colors-base.less
```diff
@@ -145,16 +148,23 @@
 @agate-heading-text-color: @agate-text-color;
 @agate-heading-accent-color: inherit;
 
+// Header
+// ---------------------------------------
+@agate-header-border-color:   inherit;
```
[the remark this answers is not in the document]
daniel-stoian-lgp: right. removed it

Input/Input.module.less
```diff
@@ -91,6 +96,23 @@
 			background-color: @agate-input-focused-background-color;
 			background-image: @agate-input-focused-background-image;
 			box-shadow: 0 0 0 @agate-input-decorator-focus-border-width @agate-input-decorator-focused-border-color;
+
+			.input {
+				&:focus {
+					opacity: @agate-input-focused-opacity;
+					color: @agate-input-text-color;
```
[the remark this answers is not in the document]
daniel-stoian-lgp: ok, i changed to @agate-input-focused-text-color, but also modified the definition of the variable for silicon say and night, as it need to be white.

0x64:
Note: Travis for the branch failed but it passed in a manual test. It seems to a happening due to mismatched `package.json`. As far as I checked, it looks okay.

Input/Input.module.less
```diff
@@ -91,6 +96,23 @@
...
+					color: @agate-input-focused-text-color;
+					caret-color: @agate-accent;
+					border-image-slice: 1;
+					border-image-source: @agate-input-active-border-image;
+					border-width: 0 0 @agate-input-focused-border-width;
+
+					&::placeholder {
+						opacity: @agate-input-opacity;
+					}
+				}
```
seunghoh, line 112: Please add this effect (change underline color) when input component get focused as well. Currently, looks like spotlight lost even the spot is on the input component. Try with up/down/left/right key to change the focus on Gallium skin

daniel-stoian-lgp: Thank you for the notice. I will update

styles/variables-silicon.less
```diff
@@ -155,7 +186,7 @@
 
 // RadioItem
 // ---------------------------------------
-@agate-radioitem-icon-border-width: 1px;
+@agate-radioitem-icon-border-width: 2px;
```
Djspaceg, line 189: If `2px` is a recurring design trend, we may need to solve this in one of several ways:
* Update the design to better support multi-resolution planning (use multiples of 3 for every value)
* Update the screenTypes.json file to more accurately reflect the final product's target display resolutions, rather than broad generalized screen resolutions, assigning a safer `pxPerRem` value with cleaner division values.
* Use a clever bit of trickery to get a clean `2px` value, putting it into a LESS variable, and using it all over this file: `~"calc(" 3px ~"- 1px)"` - This safely converts `3px` down to rem, then uses the browser to subtract `1px` for a clean `2px` end result. (The `~"..."` is LESS escaping, so LESS doesn't process the math before the conversion.)

Ex:
```css
@silicon-two-px: ~"calc(" 3px ~"- 1px)";

@agate-input-border-width: @silicon-two-px;
@agate-radioitem-icon-border-width: @silicon-two-px;
```

styles/variables-silicon.less
```diff
@@ -10,12 +10,19 @@
...
 @agate-button-badge-animation-offset: (@agate-button-animation-duration * 0.8);
 @agate-button-badge-position: -3px 0 auto auto;
-@agate-button-small-height: 48px;
-@agate-button-small-font-size: 24px;
+@agate-button-border-width: 2px;
+@agate-button-font-size: 36px;
+@agate-button-font-weight: 300;
+@agate-button-height: 75px;
+@agate-button-line-height: 1.17;
+@agate-button-border-focus-width: 0px;
```
Djspaceg, line 22: Rather than setting the border-width here, it's safer to leave the border width alone, and set the border-color to transparent instead. Border is associated with DOM layout, like padding and margin, which can cause unexpected shifting when you change it, so it's best to just avoid it altogether.

Input/Input.module.less
```diff
@@ -59,8 +63,9 @@
 
 		.input,
 		.icon {
-			line-height: @agate-input-height;
+			font-family: @agate-base-font-family;
 			height: @agate-input-height;
+			line-height: @agate-input-line-height;
```
Djspaceg, line 68: This shouldn't have needed to change. The line-height here is set to the input height to ensure that the text and icon are always properly vertically centered in the field.

The font-family should _absolutely not change_ on this line. This applies to both `.input` and `.icon`. Icon has its own defined font family. If you're intending to have the input field use a different font than the rest of the skin, is above, on line 13 or 23.

That being said, I don't think you need to actually set the font-family in this component at all. It inherits safely from the root ThemeDecorator.

Input/Input.module.less
```diff
@@ -19,12 +20,15 @@
 			text-overflow: ellipsis;
 			flex-grow: 1;
 			text-indent: 0.2em;
+			font-family: @agate-base-font-family;
```
Djspaceg, line 23: This isn't necessary.

Input/Input.module.less
```diff
@@ -10,6 +10,7 @@
 		box-shadow: 0 0 0 @agate-input-decorator-border-width @agate-input-decorator-border-color;
 		box-sizing: border-box;
 		vertical-align: middle;
+		font-family: @agate-base-font-family;
```
Djspaceg, line 13: Please only use component-based variables in module LESS files. You can assign the relationships in the variables and colors files for the skin. Referring to the general LESS values here effectively hard-codes every skin so that it _must_ use this value, rather than its own preferences.

Input/Input.module.less
```diff
@@ -91,6 +96,19 @@
 			background-color: @agate-input-focused-background-color;
 			background-image: @agate-input-focused-background-image;
 			box-shadow: 0 0 0 @agate-input-decorator-focus-border-width @agate-input-decorator-focused-border-color;
+
+			.input {
+				caret-color: @agate-accent;
```
Djspaceg, line 101: This should refer to a new `@agate-input-caret-color` variable, defined in the colors files.

In `colors-base.less`:
```css
@agate-input-caret-color: @agate-accent;
```

styles/colors-silicon-day.less
```diff
@@ -74,14 +76,17 @@
 
 // Button
 // ---------------------------------------
-@agate-button-color:          @agate-white;
-@agate-button-bg-color:       @agate-dark-gray;
+@agate-button-color:          @agate-black;
+@agate-button-bg-color:       @agate-white;
 @agate-button-bg-image:       none;
-@agate-button-border-color:   transparent;
+@agate-button-border-color:   rgba(0, 0, 0, 0.5);
```
Djspaceg, line 82: It's preferable to use LESS color functions for setting alpha values on colors:
```css
@agate-button-border-color:   fade(@agate-black, 50%);
```

styles/colors-silicon-night.less
```diff
@@ -7,6 +7,13 @@
 // as these are overrides to the "base" layer of colors.
 //
 
+// Abstracted Colors
+@agate-background: black;
+@agate-foreground: white;
+@agate-background-active: fade(white, 15%);
+@agate-foreground-active: black;
+@agate-foreground-focus: black;
```
Djspaceg, line 15: Let's refer to the `@agate` or `@silicon` color variables here instead of the hard-coded color-words.

styles/mixins.less
```diff
@@ -17,7 +17,7 @@
 .agate-item-sizing() {
 	font-size: @agate-item-font-size;
 	font-weight: @agate-item-font-weight;
-	line-height: @agate-item-height;
+	line-height: @agate-item-line-height;
```
Djspaceg, line 20: It seems counter-intuitive, but this was intentionally set as the item-height. This allows the text rendered into the item to be perfectly vertically centered, rather than potentially off-center due to an unintentional mismatch in item height and item line-height.

If your intent was to have the text not be vertically aligned in an item, then this is the correct approach; although you should restore this relationship in the base skin to not break the other skins, and set the mismatched (off-center alignment) values in the skin you want to affect.

```css
// Keep the change above
// Mixins
.agate-item-sizing() {
	...
	line-height: @agate-item-line-height;
	...
}

// Base
@agate-item-line-height: @agate-item-height;

// Silicon override
@agate-item-line-height: 300px;
```

styles/variables-base.less
```diff
@@ -258,6 +265,7 @@
 @agate-item-side-outset: @agate-component-spacing;
 @agate-item-height: 60px;
 @agate-item-height-large: 72px;
+@agate-item-line-height: 60px;
```
Djspaceg, line 268: If this is necessary, it should be a referential value like the following to be easier to work with:
```css
@agate-item-line-height: @agate-item-height;
```

styles/variables-silicon.less
```diff
@@ -10,12 +10,19 @@
 
 // Button
 // ---------------------------------------
-@agate-button-border-radius: 6px;
+@agate-button-border-radius: 37px;
```
Djspaceg, line 13: Does the border-radius really need to be exactly 37 pixels? Could this instead be a simpler or more straightforward value, like:
```css
@agate-button-border-radius: @agate-button-height;
```
or if you want to get really precise:
```css
@agate-button-border-radius: (@agate-button-height / 2);
```
`border-radius` value really only matters up to half of the height of the element when all corners are specified. So any value is ok; so we might as well set one that is easy for the browser to calculate.

Button/Button.module.less
```diff
@@ -305,4 +306,26 @@
 			}
 		});
 	}
+
+	&:global(.spottable):active,
+	:global(.spottable):active & {
```
Djspaceg, line 311: We'll likely want to make a mixin that does this, so the component LESS files can stay more focused on styling the visual states. We have a `.focus()` mixin that will be extremely similar to the implementation of this.

Also, typically we'll put state rules inside the `.applySkins()` mixin for predictability and correctness of the output selectors in complex scenarios.

Button/Button.module.less
```diff
@@ -305,4 +306,26 @@
...
+		.applySkins({
+			.bg {
+				background-color: @agate-button-active-bg-color;
+				background-image: none;
```
Djspaceg, line 315: This should be a variable, defined on base as `inherit`, and as `none` in Silicon.
Suggested name: `@agate-button-active-bg-image`

Djspaceg:
Test comment

Djspaceg:
Another test comment!

Input/Input.module.less
daniel-stoian-lgp: Good to know :)

Button/Button.module.less
daniel-stoian-lgp: Here I need some guidance. the focus() mixin comes from enact/spotlight. Should there be also an .active() mixin? What about the similar implementation, but  for focus state from this button.module.less. Should it be moved also?
I will add an agate-button-active mixin in agate mixins.less for the moment. Looking forwatd for your feedback

Djspaceg: Yes, `enact/spotlight` probably should have an `active` mixin added (at some point). Possibly even a `.visited` but the value of that may be pretty low.

Regarding the focus mixin, in this case, I don't think we actually need the elaborateness of the selector-set. One of the lines refers to any parent that has focus, that also contain a button, the other line refers to if this button is focused. We don't typically put focusable things inside focusable things, so this is probably overkill. A simpler `.focus({ rules:here; });` usage would probably be satisfactory. Also, it should live inside the `applySkins` block after the other skin rules so it can take precedence.

The new "active" mixin is pretty specific to button. We could rewrite it to be more like spotlight's existing mixins and have it accept a ruleset. Then the component (any component) can add its own rules. Here's an example (altered focus, taken from spotlight):

```css
// Focused Spottable elements
.active(@rules; @target) when (isruleset(@rules)) and (@target = parent) {
	.spottable({
		&:active {
			@rules();
		}
	}, @target);
}
.active(@rules) when (isruleset(@rules)) {
	.spottable({
		&:active {
			@rules();
		}
	});
}
```

And here's how it would be used:
```css
.button {
	...
	.applySkins({
		...

		.active({
			.bg {
				background-color: @agate-button-active-bg-color;
				background-image: @agate-button-active-bg-image;
				border: @agate-button-border-width solid @agate-button-border-color;;
			}

			&,
			.client {
				color: @agate-button-active-color;
			}

			&.highlighted {
				.bg {
					opacity: 1;
				}
			}
		});
	});
}
```

PR #257, opened 2020-05-18
Revert "FORTECH-12: Added BodyText component to Agate."

seunghoh:
Reverts enactjs/agate#251


0x64:
This is a revert PR for #251 which is merged by accident.

PR #258, opened 2020-05-18
FORTECH-12 Added BodyText component for Agate

daniel-stoian-lgp:
**Feature Added**
Added BodyText component to Agate.
Includes silicon skin styles for BodyText.


styles/mixins.less
```diff
@@ -21,3 +21,22 @@
...
+		text-decoration: none;
+	}
+	a:hover {
+		color: inherit;
+		text-decoration: none;
+	}
+	a:active {
+		color: inherit;
+		text-decoration: none;
+	}
```
Djspaceg, line 41: Let's consolidate these into a single ruleset with 4 selectors:
```css
	a:link,
	a:visited,
	a:hover,
	a:active {
		color: inherit;
		text-decoration: none;
	}
```

styles/variables-silicon.less
```diff
@@ -8,6 +8,13 @@
 // to this file with the real value for the "silicon" skin.
 //
 
+// Body
+// ----------------------------------
+@agate-body-small-font-size: 25px;
+@agate-body-font-size: 28px;
+@agate-body-font-weight: normal;
+@agate-body-line-height: 1.5;
```
Djspaceg, line 16: Please add units to line-height (`em` in this case). This value could probably be set on some skin-global variable and reused in several places, but this is the only place right now, so we can skip that piece for now if you'd like.

PR #261, opened 2020-05-19
FORTECH-17: Update styles for Checkbox

stanca-pop-lgp:
**Feature Added**
Updated Agate Checkbox component styling for gallium and silicon skins.

**Comments**
Updated Agate Checkbox component props and styling according to zeplin design (for both gallium and silicon skins).


package.json
```diff
@@ -36,10 +36,10 @@
     "root": true
   },
   "dependencies": {
-    "@enact/core": "^3.2.0",
-    "@enact/i18n": "^3.2.0",
-    "@enact/spotlight": "^3.2.0",
-    "@enact/ui": "^3.2.0",
+    "@enact/core": "^3.3.0-alpha.8",
```
daniel-stoian-lgp, line 39: please remove these changes as they are part of PR FORTECH-7

daniel-stoian-lgp: revert npm-shrinkwrap.json also

samples/sampler/package.json
```diff
@@ -24,8 +24,8 @@
   },
   "dependencies": {
     "@enact/agate": "../../",
-    "@enact/core": "^3.2.5",
-    "@enact/ui": "^3.2.5",
+    "@enact/core": "^3.3.0-alpha.8",
```
daniel-stoian-lgp, line 27: revert this change as it is part of PR FORTECH-7

package.json
stanca-pop-lgp: done

samples/sampler/package.json
stanca-pop-lgp: done

styles/variables-base.less
```diff
@@ -110,7 +110,8 @@
 
 // Checkbox
 // ---------------------------------------
-@agate-checkbox-size: 46px;
+@agate-checkbox-size: inherit;
```
Djspaceg, line 113: Often, for sizes, it's a good idea to leave some value in the base file because the variable may be used for calculations. The value "inherit" isn't a number so the compiler will panic when it attempts to do math on the string. Probably a good idea to leave this at `46px`, or at least some other number.

styles/colors-silicon-night.less
```diff
@@ -49,8 +49,9 @@
 
 // Checkbox
 // ---------------------------------------
-@agate-checkbox-text-color: @agate-white;
-@agate-checkbox-bg-color: @agate-accent;
+@agate-checkbox-text-color: @silicon-white;
```
Djspaceg, line 52: This line isn't necessary. The value is defined as white in the day version (which is inherited by this skin).

styles/colors-gallium-night.less
```diff
@@ -69,6 +69,7 @@
 // ---------------------------------------
 @agate-checkbox-text-color: @agate-white;
 @agate-checkbox-bg-color: @agate-accent;
+@agate-checkbox-border-color: transparent;
```
Djspaceg, line 72: Also not necessary, and in fact, this entire section can be deleted as it matches what's already defined in the day skin (used as the inherited base for Gallium).

ybsung:
Day mode
image
Night mode
image

The Checkbox size for night mode is smaller than the one for day mode.

ybsung:
There is no `Checkbox` ( Not `CheckboxItem` ) sampler in Agate. Could you add it in your PR by cloning it from the CheckboxItem?

styles/variables-silicon.less
```diff
@@ -21,7 +21,9 @@
 
 // Checkbox
 // ---------------------------------------
-@agate-checkbox-border-radius: 6px;
+@agate-checkbox-size: 30px;
+@agate-checkbox-border: 3px solid transparent;
```
ybsung, line 25: We already defined the border color in `@agate-checkbox-border-color: transparent;`. In addition any colors should not be defined in the variable less files. Why don't you just define a `border-width` because the default for a `border-style is a `solid`?

```
.Checkbox_toggleIcon__3FAw8.silicon {
    width: 1.25rem;
    height: 1.25rem;
    border: 0.125rem solid transparent; // This border color will be overwritten by the following border-color.
    border-color: transparent;
    border-radius: 0;
    color: #ffffff;
    background-color: #f1304f;
}
```

ybsung:
Hello @stanca-pop-lgp 
It looks much better to me.

I found one issue.
Day mode
image
Night mode
image
In night mode, the text color for the title - `Agate Checkbox` is black and I can't see it because the background color is also black.

styles/variables-gallium.less
```diff
@@ -14,6 +14,8 @@
 
 // Checkbox
 // ---------------------------------------
+@agate-checkbox-size: 46px;
```
ybsung, line 17: Do we need to define it here because it is already defined with the same value in styles/variables-base.less (@agate-checkbox-size: 46px;)?

stanca-pop-lgp: No, I forgot to delete this. Thank you.

stanca-pop-lgp:
> Hello @stanca-pop-lgp
> It looks much better to me.
> 
> I found one issue.
> Day mode
> image
> Night mode
> image
> In night mode, the text color for the title - `Agate Checkbox` is black and I can't see it because the background color is also black.

This is because the Agate storybook doesn't have styles for silicon night mode. So you couldn't have seen the checkbox if I had left it as it is. I just replaced the background color for now, but we will update all the variables for all the components in order to be visible in night mode. Daniel already did some changes here and pushed them, that is why I didn't intervene so much.

ybsung:
@stanca-pop-lgp 
Thanks for adding the unit test. LGTM.

styles/colors-silicon-day.less
```diff
@@ -92,8 +92,10 @@
 
 // Checkbox
 // ---------------------------------------
-@agate-checkbox-text-color: @agate-white;
-@agate-checkbox-bg-color: @agate-accent;
+@agate-checkbox-text-color: @agate-black;
+@agate-checkbox-bg-color: transparent;
+@agate-checkbox-border-color: rgba(0, 0, 0, 0.5);
```
Djspaceg, line 97: Let's use the `fade` LESS color function here: `fade(@agate-black, 50%)`

styles/colors-silicon-night.less
```diff
@@ -11,8 +11,9 @@
 @agate-bg-color: black;
 @agate-bg-image: none;
 @agate-bg-image2: none;
-@agate-text-color: @agate-foreground;
-@agate-title-text-color: @agate-foreground;
+@agate-text-color: @agate-background;
+@agate-title-text-color: @agate-background;
+@agate-background: @agate-foreground;
```
Djspaceg, line 16: This doesn't seem right. I think it may make more sense to update these (foreground, background) general color assignments, rather than reverse where they are assigned.

If they're updated, many other values will automatically pick up the correct colors and work on other components will become simpler, since you're establishing the relationships between the variables and ultimately their color values.

```css
@agate-background: @silicon-black;
@agate-foreground: @silicon-white;
```

The `@silicon` colors are defined in the main Silicon skin colors file (day), which is the basis for this file. All variables defined there are accessible here.

Since the relationship between the conceptual variable `@agate-text-color` and the conceptual color `@agate-foreground` is already established in the other (day) file, you actually don't even need to recreate that rule here, even though the values are updated in this file. Each of these files can be considered to have their own scope.

styles/colors-silicon-night.less
```diff
@@ -50,7 +51,8 @@
 // Checkbox
 // ---------------------------------------
 @agate-checkbox-text-color: @agate-white;
-@agate-checkbox-bg-color: @agate-accent;
+@agate-checkbox-border-color: rgba(255, 255, 255, 0.5);
```
Djspaceg, line 54: `fade` function here too: `fade(@agate-foreground, 50%)` (will work correctly after changing the definition above). In fact, you could set this `fade(@agate-foreground, 50%)` rule in the day file (replacing the checkbox border color line) and omit it from here, since the relationship is the same in night mode. Neato.

Now that I look more deeply, updating the @agate-foreground rule mentioned above would remove the need to set all of these checkbox values, since they can just use the conceptual color values.

Djspaceg:
I went ahead and resolved the merge conflict after `FORTECH-26-27-28-30` was merged.

Reminder: Please reassign the JIRA ticket back to me after you're finished so I know the PR is ready for integration

PR #263, opened 2020-05-20
PLAT-107712: Fix `inline` not found error on LabeledIcon and LabeledIconButton

seunghoh:
Fixed "inline is not a function" error on LabeledIcon and LabeledIconButton.
This error is due to ui/Slottable migration from `kind` to `function`
Changed import from `ui/LabeledIcon` to `ui/LabeldIconBase` as it looks make sense to me rather than removing the use of inline

### DIFF
--- a/LabeledIcon/LabeledIcon.js
+++ b/LabeledIcon/LabeledIcon.js
@@ -11,7 +11,7 @@
  */
 
 import kind from '@enact/core/kind';
-import UiLabeledIcon from '@enact/ui/LabeledIcon';
+import {LabeledIconBase as UiLabeledIcon} from '@enact/ui/LabeledIcon';
 import Pure from '@enact/ui/internal/Pure';
 import compose from 'ramda/src/compose';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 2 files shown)


LabeledIcon/LabeledIcon.js
```diff
@@ -11,7 +11,7 @@
  */
 
 import kind from '@enact/core/kind';
-import UiLabeledIcon from '@enact/ui/LabeledIcon';
+import {LabeledIconBase as UiLabeledIcon} from '@enact/ui/LabeledIcon';
```
ybsung, line 14: Let's rename `LabeledIconBase` as `UiLabeledIconBase` and wrap `agate/LabeledIconBase` with `ui/LabeledIconDecorator` like `agate/Button`.

```
import {LabeledIconBase as UiLabeledIconBase, LabeledIconDecorator as UiLabeledIconDecorator} from '@enact/ui/LabeledIcon';

...

 * @hoc
 * @memberof agate/LabeledIcon
 * @mixes agate/LabeledIcon.UiLabeledIconDecorator
 * @mixes ui/Pure.Pure
 * @public
 */
const LabeledIconDecorator = compose(
	UiLabeledIconDecorator,
	Pure,
	Skinnable
);
```

PR #264, opened 2020-05-22
PLAT-106938: Add prop to PopupMenu for customizing closeButton string 

seunghoh:
Add `closeButtonLabel` prop so that developer can customize string for `closeButton`
Add `popupMenu` sampler

### DIFF
--- a/samples/sampler/stories/default/PopupMenu.js
+++ b/samples/sampler/stories/default/PopupMenu.js
@@ -5,9 +5,9 @@ import React from 'react';
 import {storiesOf} from '@storybook/react';
 
 import LabeledIconButton from '@enact/agate/LabeledIconButton';
-import PopupMenu from '@enact/agate/PopupMenu';
+import {PopupMenu, PopupMenuBase} from '@enact/agate/PopupMenu';
 
-const Config = mergeComponentMetadata('PopupMenu', PopupMenu);
+const Config = mergeComponentMetadata('PopupMenu', PopupMenuBase);
 
 storiesOf('Agate', module)
 	.add(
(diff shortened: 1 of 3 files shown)


samples/sampler/stories/default/PopupMenu.js
```diff
@@ -0,0 +1,43 @@
+import {mergeComponentMetadata} from '@enact/storybook-utils';
+import {action} from '@enact/storybook-utils/addons/actions';
+import {boolean, select, text} from '@enact/storybook-utils/addons/knobs';
+import React from 'react';
+import {storiesOf} from '@storybook/react';
+
+import LabeledIconButton from '@enact/agate/LabeledIconButton';
+import PopupMenu from '@enact/agate/PopupMenu';
+
+const Config = mergeComponentMetadata('PopupMenu', PopupMenu);
```
ybsung, line 10: We need the `PopupMenuBase` instead of the `PopupMenu` here because all prop types and default props are defined not in the `PopupMenuBase` but in the `PopupMenu`.

```
import {PopupMenu, PopupMenuBase} from '@enact/agate/PopupMenu';

const Config = mergeComponentMetadata('PopupMenu', PopupMenuBase);
```

samples/sampler/stories/default/PopupMenu.js
```diff
@@ -0,0 +1,43 @@
...
+
+storiesOf('Agate', module)
+	.add(
+		'PopupMenu',
+		() => (
+			<div>
+				<PopupMenu
+					closeButton={boolean('closeButton', Config)}
+					closeButtonLabel={text('closeButtonLabel', Config, 'Cancel')}
```
ybsung, line 19: If using `PopupMenuBase`, then we could omit the third parameter - `Cancel`.

samples/sampler/stories/default/PopupMenu.js
```diff
@@ -0,0 +1,43 @@
...
+					noAnimation={boolean('noAnimation', Config)}
+					noAutoDismiss={boolean('noAutoDismiss', Config)}
+					onClose={action('onClose')}
+					onHide={action('onHide')}
+					open={boolean('open', Config)}
+					orientation={select('orientation', ['horizontal'], Config, 'horizontal')}
```
ybsung, line 25: Same above. Please omit the `horizontal`.

samples/sampler/stories/default/PopupMenu.js
seunghoh: Done, Thanks!
I also fixed Popup sampler

ybsung:
LGTM

PR #269, opened 2020-06-09
PLAT-104180: VirtualList: Fix jumping focus when keys pressed in a row

MikyungKim:
This PR fixes jumping focus when keys pressed in a row in VirtualList.
1) directional keys right after Page up/down
2) an opposite directioanl key right after a series of directional keys

The first issue occurs when a user pressed directional key(up/down) right after page up/down key.
The list blurs the current focus and scrolls to the target position by page up/down key and then focuses the item. But if the directional key got in before focusing the item by page up/down key, the spotlight gives the focus the item nearby mouse pointer position(https://github.com/enactjs/enact/pull/1292), and then list focuses the item by page up/down key.
I confirmed that this issue is from moonstone and fixed not to focus when there is a focus already on the list and save the target scroll position(https://github.com/enactjs/enact/pull/2778) to make the spotlight focused item scroll into view.

The second issue occurs when a user pressed a series of directional keys and pressed the opposite side of the key.
The main root cause was that the list sticks to the 'start' and 'end' based on `index` and `nextIndex`. So if `nextIndex` is smaller than current `index` meaning backward direction, list always sticks to 'start' which is wrong because we don't need this kind of jump.
So I changed the condition to stick to the nearest one between 'start' and 'end'. Also, I had to clear `preservedIndex` set by previous key handling not to focus after we found the target in the current viewport.

### DIFF
--- a/VirtualList/useThemeVirtualList.js
+++ b/VirtualList/useThemeVirtualList.js
@@ -147,6 +147,7 @@ const useSpottable = (props, instances) => {
 				focusByIndex(nextIndex);
 			} else {
 				const itemNode = getItemNode(nextIndex);
+				const stickTo = Math.abs(endBoundary - end) < Math.abs(startBoundary - start) ? 'end' : 'start';
 
 				mutableRef.current.isScrolledBy5way = true;
 				mutableRef.current.isWrappedBy5way = isWrapped;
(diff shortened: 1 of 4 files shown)


MikyungKim:
Closing this in favor of https://github.com/enyojs/sandstone/pull/440

PR #272, opened 2020-06-15
FORTECH-44 Add RangePicker to Agate

daniel-stoian-lgp:
**Feature Added**
Added new component RangePicker
Moved common logic with the Picker component to an internal component


internal/Picker/tests/Picker-specs.js
```diff
@@ -0,0 +1,123 @@
...
+
+		increment(picker);
+
+		const expected = 0;
+		const actual = handleChange.mock.calls.length;
+
+		expect(actual).toBe(expected);
+	});
+
+
```
stanca-pop-lgp, line 68: Did you leave 3 empty lines intentionally?

styles/colors-gallium-day.less
```diff
@@ -149,10 +149,10 @@
 @agate-picker-active-color: inherit;
 @agate-picker-active-bg-color: @agate-accent;
 @agate-picker-active-bg-image: @agate-gradient;
-@agate-picker-item-top-bg-color: none;
```
stanca-pop-lgp, line 152: You may want to use "transparent" instead of "none" for background-color. "none" is not a valid value for the background-color property.

daniel-stoian-lgp:
1. I put a question on Zeplin regarding this matter
2. Solved
3. If the others agree with this, I  don't have a problem applying it

daniel-stoian-lgp:
> 1. Don't we need to support "wheel up/down" and "drag up/down"? From the comment, From the GUI comment it ways `The operation of the Picker is the same as the iOS Date Picker.`
> 2. GUI guide shows upper 2 items and lower 2 items. How can we show?

Hello, Yes, this will need to support "wheel up/down" and "drag up/down" in the end.
BUT, I put the question on the GUI egarding the transition after I gave this PR for review.
The comment flow on the GUI is not finished yet, some other screens being expected to be sent.
At this moment, I want to receive feedback regarding the implementation without "wheel up/down" and "drag up/down" functionality because The PR is quite big as it is.

ybsung:
Hello @daniel-stoian-lgp 

There are some visual regression which break some skins except silicon.

### DateTimePicker

- Agate Gallium

Diff
image

develop Sampler
image

your branch Sampler
image

ybsung:
### Checklist

npm run parse - Failed

- [ ] Known parse errors: ThumbnailItem, useChainRefs
- [ ] Invalid reference: agate/Picker.PickerDecorator: type: mixes - Picker in /Users/yb.sung/Documents/_Work/docs/raw/agate/Picker/Picker.js:128


npm run test-ss - Failed

- [ ] DateTimePicker - Agate Gallium

npm run test-ui - Passed

ybsung:
>  Invalid reference: agate/Picker.PickerDecorator: type: mixes - Picker in /Users/yb.sung/Documents/_Work/docs/raw/agate/Picker/Picker.js:128

Please check my commit to fix js doc parse errors.
https://github.com/enactjs/agate/commit/1ad4ba6a801eabab73cc5489fa9bdab4c693b930

daniel-stoian-lgp:
> Hello @daniel-stoian-lgp
> 
> There are some visual regression which break some skins except silicon.
> 
> ### DateTimePicker
> * Agate Gallium
> 
> Diff
> image
> 
> develop Sampler
> image
> 
> your branch Sampler
> image

Reverted to the original colors

ybsung:
Remark:
```
Module: =====================
Component: -----------------------------
⇧: HoC
↓: Composition
+: public prop
-: private prop
```

### Before PR

**Picker module**
```
=====================
Picker module
=====================
+ Picker
+ PickerBase
+ PickerDecorator
=====================
```

**Picker**
```
IdProvider, Changeable, Skinnable
⇧
-----------------------------
PickerDecorator
-----------------------------
@ mixes IdProvider
@ mixes Changeable
@ mixes Skinnable
-----------------------------
⇧
-----------------------------
Picker
-----------------------------
↓
-----------------------------
PickerBase
-----------------------------
+ children
+ 'aria-valuetext'
+ decrementAriaLabel
- id
+ incrementAriaLabel
+ value: 0
-----------------------------
↓
div
↓
child ( from props.children )
```

### After PR

**Picker module**
```
=====================
Picker module
=====================
+ Picker
+ PickerBase
+ PickerDecorator
=====================
```

**Picker**
```
IdProvider, Changeable, Skinnable
⇧
-----------------------------
PickerDecorator = internal/Picker.PickerDecorator
-----------------------------
⇧
-----------------------------
Picker
-----------------------------
↓
-----------------------------
PickerBase
-----------------------------
+ children
+ disabled
+ onChange
+ orientation
+ value: 0
-----------------------------
↓
internal/Picker 
↓
PickerItem
↓
Marquee
↓
div
↓
child ( from props.children )
```

**internal/Picker module**
```
=====================
internal/Picker module
=====================
+ Picker
+ PickerDecorator
=====================
```

**internal/Picker.Picker**
```
-----------------------------
internal/Picker.PickerDecorator
-----------------------------
@ mixes IdProvider
@ mixes Changeable
@ mixes Skinnable
-----------------------------
⇧
-----------------------------
internal/Picker.Picker
-----------------------------
↓
-----------------------------
internal/Picker.PickerBase
-----------------------------
+ index
+ max
+ min
+ 'aria-valuetext'
+ children
+ className
+ decrementAriaLabel
+ disabled
- id
+ incrementAriaLabel
+ onChange
+ orientation: 'vertical'
+ step: 1
+ value: 0
-----------------------------
↓
PickerViewManager
```

ybsung:
Hello @daniel-stoian-lgp 

Before reviewing your code in detail,
I'd like to share some comments because your patch affects other skins.

1. We hope that we do not break any APIs until we make a decision.

In agate/Picker.Picker,

`aria-valuetext`, `decrementAriaLabel`, and `incrementAriaLabel` are removed from the its propTypes. Pickers might work properly because those props are handled in agate/internal/Picker.Picker. But app developers could think that some APIs are removed. In this case, please leave the related comments.

If those props are not handled in the agate/Picker.Picker, then we could define them out of it.

```
const PickerBase = kind({
	name: 'Picker',

	propTypes: /** @lends agate/internal/Picker.Picker.prototype */ {
	...
});

/**
 * Sets the hint string read when focusing the decrement button.
 *
 * name decrementAriaLabel
 * @memberof agate/Picker.Picker.prototype
 * @default 'previous item'
 * @type {String}
 * @public
 */
```

agate/RangePicker.RangePicker's `orientation` prop is also not handled in the RangePicker. So you could define its JSDoc out of the RangePicker.

2. Please preserve the previous behavior in other skins.
The child from the props.children was wrapped with just `div`. But now it is wrapped with Marquee. I don't know that we should apply Marquee to all Picker's content for all skins.

3. internal/Picker becomes heavy including `PickerViewManager`.
I think that it might be better to provide any type so that some special logics could be applied depending on the type.

internal/Picker/Picker.js
```diff
@@ -0,0 +1,338 @@
...
+		);
+	}
+});
+
+/**
+ * Applies Agate specific behaviors to [PickerBase]{@link agate/Picker.PickerBase}.
+ *
+ * @hoc
+ * @memberof agate/internal/Picker
+ * @mixes ui/Changeable.Changeable
```
ybsung, line 321: Please add JS doc for `IdProvider`.

ybsung:
@daniel-stoian-lgp 

I know that we need to change lots of tests and jsdocs when changing component hierarchy. For me as a reviewer, it's fine to defer to update them until the hierarchy has been concreated.

Picker/Picker.js
```diff
@@ -116,108 +72,71 @@ const PickerBase = kind({
...
-			handleFlick,
-			handleIncrement,
-			incrementAriaLabel,
-			value,
-			valueId,
-			...rest
-		} = props;
-		const isFirst = value <= 0;
-		const isLast = value >= React.Children.count(values) - 1;
+		const {children, value, max, ...rest} = props;
```
ybsung, line 85: We usually define `prop`s alphabetically like `{children, max, value, ...rest}`.

ybsung:
Depending on the picker item content, the picker item width changes. It might not be changed.

image

image

ybsung:
When launching the RangePicker sampler, the content with `4` has focused even though not hovering on it.

image

samples/sampler/stories/default/RangePicker.js
```diff
@@ -0,0 +1,30 @@
...
+		'RangePicker',
+		() => (
+			<div style={{padding: '0 20%'}}>
+				<RangePicker
+					disabled={boolean('disabled', Config)}
+					onChange={action('onChange')}
+					min={number('min', Config, 0)}
+					max={number('max', Config, 20)}
+					step={number('step', Config, 1)}
+					defaultValue={number('defaultValue', Config, 5)}
```
ybsung, line 22: Even though changing the `defautValue` value in the Knobs, it didn't apply it to the RangePicker because the RangePicker already rendered. One possible solution is the one adding the `key={number('defaultValue', Config, 5)}` prop in the `RangePicker`. Then the RangePicker will render again when the `defaultValue` changes.

ybsung: Please write props alphabetically. So the `defaultValue` would be the first prop in the RangePicker.

internal/Picker/Picker.js
```diff
@@ -0,0 +1,316 @@
...
+
+		return (
+			<PickerRoot {...rest} onFlick={handleFlick}>
+				<PickerButtonItem
+					aria-controls={valueId}
+					aria-disabled={isFirst}
+					aria-label={decrementAriaLabel}
+					className={css.itemDecrement}
+					onClick={handleDecrement}
+					disabled={isFirst}
```
ybsung, line 261: Please sort the props alphabetically.

internal/Picker/Picker.js
```diff
@@ -0,0 +1,316 @@
...
+				>
+					<div className={css.label}>
+						{isFirst ? '' :  clamp(min, max, Array.isArray(values) ? values[value - step] : value - step)}
```
ybsung, line 264: The `clamp(min, max, Array.isArray(values)` log is used twice and i defined in JSX. It would be better to define it as a variable out of JSX, but in render function.

RangePicker/RangePicker.js
```diff
@@ -0,0 +1,202 @@
...
+/**
+ * A component that lets the user select a number from a range of numbers.
+ *
+ * By default, `RangePicker` maintains the state of its `value` property. Supply the `defaultValue`
+ * property to control its initial value. If you wish to directly control updates to the component,
+ * supply a value to `value` at creation time and update it in response to `onChange` events.
+ *
+ * @class RangePicker
+ * @memberof agate/RangePicker
+ * @mixes ui/Changeable.Changeable
```
ybsung, line 182: Please add `mixes` for Pure.

internal/Picker/Picker.module.less
```diff
@@ -52,14 +58,32 @@
...
+
+			.active,
+			.itemDecrement,
+			.itemIncrement{
+				padding: @agate-picker-item-padding;
+			}
 		}
 	});
-}
+}
```
ybsung, line 89: Please add one empty line at the end of the file.

styles/variables-base.less
```diff
@@ -318,12 +318,17 @@
 @agate-panels-vertical-tab-margin: 0;
 @agate-panels-vertical-tab-width: inherit;
 
+// Picker
+// ---------------------------------------
 @agate-picker-border-radius: 0;
+@agate-picker-item-max-width: 300px;
+@agate-picker-item-padding: 0 24px;
```
ybsung, line 325: I'm not sure that it's fine to define silicon skin's picker item padding in `variables-base.less`.
How about defining `@agate-picker-item-padding` with `0` in `variables-base.less` and with `0 24px` in `variables-silicon.less`?

styles/variables-silicon.less
```diff
@@ -127,7 +127,9 @@
 
 // Picker
 // ---------------------------------------
-@agate-picker-font-weight: 500;
+@agate-picker-font-size: 28px;
```
ybsung, line 130: If possible, we want to use a multiple of 3. Please check https://github.com/enactjs/agate/pull/253#discussion_r431506007 discussion.

samples/sampler/stories/default/RangePicker.js
webOS101: `defaultValue` shouldn't be used in samplers.   Use `value` and control it or don't modify the default value for already created controls.

samples/sampler/stories/default/RangePicker.js
```diff
@@ -0,0 +1,30 @@
...
+	.add(
+		'RangePicker',
+		() => (
+			<div style={{padding: '0 20%'}}>
+				<RangePicker
+					disabled={boolean('disabled', Config)}
+					onChange={action('onChange')}
+					min={number('min', Config, 0)}
+					max={number('max', Config, 20)}
+					step={number('step', Config, 1)}
```
ybsung, line 21: Please do not define any default value of the boolean, number, select in '@enact/storybook-utils/addons/knobs' if that default value is already defined in our component.

The `step` and the `orientation` are defined in `@agate/internal/Picker`. So here is the way that we apply those values to the `number` and the `select`.

1. Import modules having the defaultProps
2. Append those modules as the parameters of the mergeComponentMetadata function.
3. Omit the default values in number, select, and so on.

Then `Config` will pass the component's default value to boolean, number, and select knobs.

```
import {PickerBase} from '@enact/agate/internal/Picker';

const Config = mergeComponentMetadata('RangePicker', RangePicker, PickerBase);

<RangePicker
	defaultValue={number('defaultValue', Config, 5)}
	key={number('defaultValue', Config, 5)
	onChange={action('onChange')}
	max={number('max', Config, 20)}
	min={number('min', Config, 0)}
	orientation={select2('orientation', ['vertical', 'horizontal'], Config)}
	step={number('step', Config)}
/>
```

samples/sampler/stories/default/Picker.js
```diff
@@ -16,6 +16,7 @@ storiesOf('Agate', module)
 				<Picker
 					disabled={boolean('disabled', Config)}
 					onChange={action('onChange')}
+					orientation={select('orientation', ['vertical', 'horizontal'], Config)}
```
ybsung, line 19: The `Config` does not have the default value for the `orientation`. Please check the comment below for `RangePicker` sampler - https://github.com/enactjs/agate/pull/272/files#r444642557.

samples/sampler/stories/default/RangePicker.js
ybsung: Oh, you have to export `PickerBase` in internal/Picker first.

daniel-stoian-lgp: Thanks for the tip. I was wondering how the defaultValues are brought in this case

ybsung: https://github.com/enactjs/agate/pull/272#discussion_r444638206

As roy mentioned, the defaultValue shouldn't be used in samplers.

internal/Picker/Picker.js
daniel-stoian-lgp: ok, i will define currentValue, decrementValue and incrementValue inside render()

ybsung:
I fixed minor things.
- Reroder exported modules internal/Picker/Picker.js.
- Remove the step default value, paddings in RangePicker sampler.
- Define defaultValue in RangePicker sampler. If not, there is a warning because the `value` prop is required As roy mentioned, I don't allow to modify the default value.
> defaultValue shouldn't be used in samplers. Use value and control it or don't modify the default value for already created controls.

Djspaceg:
@daniel-stoian-lgp, this PR was complaining about the CLA. I know you've set it up in the past, because this is the first time it's showed up, so maybe it needs to be re-issued? We're ok now, but if you could look into it, that'll help us out. Thanks!

PR #274, opened 2020-06-17
FORTECH-50: Silicon night colors and variables clean-up

stanca-pop-lgp:
Feature Added
Refactored silicon and gallium day and night colors files.

Comments
Removed duplicates from color less files, considering the inheritance rule: colors-base -> colors-day > colors-night.

### DIFF
--- a/styles/colors-carbon.less
+++ b/styles/colors-carbon.less
@@ -104,7 +104,7 @@
 @agate-item-focus-bg-color: @agate-spotlight-bg-color;
 @agate-item-focus-color: @agate-spotlight-color;
 @agate-item-selected-color: @agate-item-label-color;
-@agate-item-selected-outline: 5px solid @agate-item-focus-bg-color;
+@agate-item-selected-outline-color: 5px solid @agate-item-focus-bg-color;
 
 // LabeledIconButton
 // ---------------------------------------
--- a/styles/colors-carbon.less
+++ b/styles/colors-carbon.less
@@ -186,8 +186,8 @@
 @agate-slider-knob-bg-image:  linear-gradient(to bottom, transparent, fade(#333, 50%));
 @agate-slider-knob-focus-bg-color: @agate-slider-knob-bg-color;
 @agate-slider-knob-focus-bg-image: @agate-slider-knob-bg-image;
-@agate-slider-knob-shadow:    0 3px 9px 3px fade(black, 30%);
-@agate-slider-knob-focus-shadow: @agate-slider-knob-shadow;
+@agate-slider-knob-shadow-color: fade(black, 30%);
+@agate-slider-knob-focus-shadow-color: @agate-slider-knob-shadow-color;
 
 // IncrementSlider
 // ---------------------------------------
@@ -227,8 +227,8 @@
 @agate-sliderbutton-focus-bg-image: @agate-slider-track-focus-bg-image;
 @agate-sliderbutton-knob-bg-color:  @agate-slider-knob-bg-color;
 @agate-sliderbutton-knob-bg-image:  @agate-slider-knob-bg-image;
-@agate-sliderbutton-knob-shadow:    @agate-slider-knob-shadow;
-@agate-sliderbutton-knob-focus-shadow: @agate-slider-knob-focus-shadow;
+@agate-sliderbutton-knob-shadow-color:    @agate-slider-knob-shadow-color;
+@agate-sliderbutton-knob-focus-shadow-color: @agate-slider-knob-focus-shadow-color;
 
 // Spinner
 // ---------------------------------------
--- a/styles/colors-carbon.less
+++ b/styles/colors-carbon.less
@@ -42,7 +42,7 @@
 // ---------------------------------------
 @agate-button-color:          @agate-foreground;
 @agate-button-bg-color:       @agate-background;
-@agate-button-bg-image:       linear-gradient(to bottom left, transparent, #333);
+@agate-button-bg-image:       linear-gradient(to bottom left, transparent, @agate-darker-gray);
 @agate-button-border-color:   @agate-accent;
 @agate-button-shadow:         none;
 @agate-button-focus-color:    @agate-spotlight-color;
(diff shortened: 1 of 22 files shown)


ybsung:
Hello @stanca-pop-lgp 

Very great job!!! Thanks for your effort.

I found one issue. So I'd like to introduce it.
Have you ever executed Enact screenshot test in agate?

If no, here is the steps how to use it.
1) Execute `npm run test-ss` in your agate root directory with `develop` branch first.
2) Execute it again with your branch - `feature/FORTECH-50`.
3) Then screentshot test will produce the screenshot diff images if there are some differences of agate components between `develop` and your branch.

For example, I found one change in your branch from the `develop` branch.
In Gallium day mode, the border color of the Popup and the text color in the Popup changed. Here is the diff image generated from `npm run test--ss` in `⁨agate⁩ ▸ ⁨tests⁩ ▸ ⁨screenshot⁩ ▸ ⁨dist⁩ ▸ ⁨screenshots⁩ ▸ ⁨diff⁩` directory. They shouldn't be changed.
image

When running agate sampler, you could confirm it.
- With `develop` branch
image

- With `feature/FORTECH-50` branch
image

ybsung:
https://github.com/enactjs/agate/blob/fdd341cd93d0cf5e7e73ccbea1bb1fd9824deede/styles/colors-gallium-day.less#L112

I think it might be better to define a less variable for `5x`.
In addition, I hope that we could use a multiple of 3 instead of `5px`.
For more information, please check https://github.com/enactjs/agate/pull/253#discussion_r431506007 or `Tip 10>` that I mentioned in slack.

https://github.com/enactjs/agate/blob/fdd341cd93d0cf5e7e73ccbea1bb1fd9824deede/styles/colors-gallium-day.less#L129

`@agate-white` instead of `white`

https://github.com/enactjs/agate/blob/fdd341cd93d0cf5e7e73ccbea1bb1fd9824deede/styles/colors-gallium-day.less#L166
```
@agate-progressbar-fill-bg-image: linear-gradient(to right, transparent, fade(white, 75%));
@agate-progressbar-vertical-fill-bg-image: linear-gradient(to top, transparent, fade(white, 75%));
```
`@agate-white` instead of `white`

https://github.com/enactjs/agate/blob/fdd341cd93d0cf5e7e73ccbea1bb1fd9824deede/styles/colors-gallium-day.less#L184

Let's define a less variable for `0 3px 9px 3px`.

https://github.com/enactjs/agate/blob/fdd341cd93d0cf5e7e73ccbea1bb1fd9824deede/styles/colors-gallium-day.less#L219

https://github.com/enactjs/agate/blob/fdd341cd93d0cf5e7e73ccbea1bb1fd9824deede/styles/colors-gallium-day.less#L224

`@agate-black` instead of `black`

https://github.com/enactjs/agate/blob/fdd341cd93d0cf5e7e73ccbea1bb1fd9824deede/styles/colors-gallium-day.less#L234

Let's define `#6c6c6c` as a less variable.

styles/colors-carbon.less
```diff
@@ -186,8 +186,8 @@
 @agate-slider-knob-bg-image:  linear-gradient(to bottom, transparent, fade(#333, 50%));
```
ybsung, line 186: Let's replace #333 with @agate-darker-gray.

ybsung: I found #333 more `4 times in colors-carbon.less` and `2 times in colors-silicon-day.less`.

styles/variables-base.less
```diff
@@ -277,6 +277,7 @@
 @agate-item-line-height-large: @agate-item-height-large - (@agate-item-vertical-padding * 2);
 @agate-item-indent: 36px;
 @agate-item-border-width: inherit;
+@agate-item-selected-outline-width: 6px;
```
ybsung, line 280: This variable might not be used anywhere.

ybsung: I fixed it.

ybsung:
Hello @stanca-pop-lgp 

I fixed some screenshot test fails in f74f69f commit.
It seems to me that the left fails that I didn't fix are not bugs because the background color changes from black to `#1f1f1f` and it's the correct fixes. So let's go ahead.

```
 Agate Gallium Night
     Button
        ✖ Button~/Agate Gallium Night~/click me
        ✖ Button~/Agate Gallium Night~/Lorem ipsum d…, selected = true
        ✖ Button~/Agate Gallium Night~/Not Selected.…

     LabeledIconButton
        ✖ LabeledIconButton~/Agate Gallium Night~/default

     ToggleButton
        ✖ ToggleButton~/Agate Gallium Night~/default
```

ybsung:
@stanca-pop-lgp and I have the following concerns. 

- `#6c6c6c` color is now used directly instead of using a less variable. What is a proper variable name for it?
- The `black` color value couldn't be replace with `@agate-black` because `@agate-black` is not actually `black` like `@agate-black: #131313` in gallium-day skin. Is it fine to keep using `black` string as a color value or is there any better way for it?

I left some comments to hear a merger's opinion. CC: @Djspaceg 

Others looks great to me. Thanks @stanca-pop-lgp  for your effort.

Djspaceg:
> @stanca-pop-lgp and I have the following concerns.
> 
> * `#6c6c6c` color is now used directly instead of using a less variable. What is a proper variable name for it?
> * The `black` color value couldn't be replace with `@agate-black` because `@agate-black` is not actually `black` like `@agate-black: #131313` in gallium-day skin. Is it fine to keep using `black` string as a color value or is there any better way for it?

`black` is ok to refer to directly here. This is actually exactly why we define `@agate-black`, so it can be different but still human-readable as "black".

As for `#6c6c6c`, that's pretty middle-of-the-road gray to me. I don't have a strong preference. Sometimes it's ok to just refer to the color value directly. It's subjective about what the author thinks is best.

Djspaceg:
This looks like it was a huge effort. If I understand the enormous change-set (+1,154 −1,652, wow), I'd break it down into 3 categories.

1. Simplification and redundancy removal of the Gallium and Silicon night skins
2. Indenting/formatting of variables
3. Updating of values to match the guides
4. Changes to shadow variables

Questions/Followup
1. How did you validate that these were redundant? Was It manual or were you able to do a search or something? Looks like a huge amount of greatly needed work.
Did I share the Perl script I wrote to help discover unused LESS variables? It might need a little tuning to work well with Agate.
2. While I applaud and appreciate the uniformity of the way the files look now, this presents a pretty large problem in terms of merge conflicts with other PRs and will cause a significant amount of white-space churn when a long variable is added or removed. These types of changes make it difficult to recognize when or if values change intentionally or unintentionally as well. Typically, we'd reserve this sort of work to be done directly after a release, before new work has been started, and try to get it merged ASAP to reduce the problem as much as possible. I'm hesitant to say "please revert it" because I do appreciate the work, but at the same time, I'm hesitant to move it forward, recognizing the amount of follow-up headache this will produce. Maybe the pending PRs can be rebased onto this branch to help resolve that conflict?
3. Super! Glad this was done. However, because of the above, it makes it really difficult to follow what was changed on purpose. Looks like @ybsung has helped sort out some with Screenshot testing.
4. Shadows are sort of a special case. They are multi-valued, but not true "CSS shorthand", like `margin`, `padding`, or even `background`; they don't have separate rules for each of the parts and must always be treated as a whole when assigning values. For this reason, we decided to keep the measurements and the color together and define the single variable once in the colors files. I know it looks awkward having numbers in the colors file, but it's a small exception. From a skinning perspective, it may make sense for a skin to define a different blur or shadow offset distance in a light skin or a dark skin, depending on what suits the background and other colors better. Having the measurements in the variables file removes that capability. From a technical perspective, having it broken into two variables produces the following result if only one or none is defined: `box-shadow: inherit black;` or `box-shadow: inherit inherit;`, respectively. Please switch the variables back to a single shadow variable form.

PR #277, opened 2020-06-22
FORTECH-64 Add silicon styles for Slider

daniel-stoian-lgp:
Feature Added
Added silicon styles for Slider

### DIFF
--- a/styles/variables-carbon.less
+++ b/styles/variables-carbon.less
@@ -78,7 +78,7 @@
 
 // Slider
 // ---------------------------------------
-@agate-slider-knob-active-transform: @agate-translate-center scale(1);
+@agate-slider-active-knob-transform: @agate-translate-center scale(1);
 @agate-slider-knob-border-radius: (@agate-slider-knob-height / 2);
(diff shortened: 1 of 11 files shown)


ybsung:
Hello @daniel-stoian-lgp 

Did you see the following screenshot? The enlarged knob is clipped in the sampler.

image

styles/variables-carbon.less
```diff
@@ -78,7 +78,7 @@
 
 // Slider
 // ---------------------------------------
-@agate-slider-knob-active-transform: @agate-translate-center scale(1);
+@agate-slider-active-knob-transform: @agate-translate-center scale(1);
```
ybsung, line 81: Could we omit `scale(1)`? In my opinion, by default, the scale would be `1`. Please fix me if I'm wrong.

ybsung: If we have to use the `scale`, please use scale`(@agate-slider-knob-resting-state-scale)`

daniel-stoian-lgp: I removed the scale. it was not necessary

PR #276, opened 2020-06-18
FORTECH-49 Update Input

adrian-cocoara-lgp:
Updated silicon styles for Input 

Enact-DCO-1.0-Signed-off-by: Adrian Cocoara adrian.cocoara@lgepartner.com

PR #279, opened 2020-06-29
FORTECH-65: Add OSD mode for Popup

daniel-stoian-lgp:
Feature Added
Added new property for Popup whihc makes it appear on the top of the screen and in full wdth


styles/colors-silicon-day.less
```diff
@@ -52,7 +52,7 @@
 @agate-darker-gray: @silicon-dark;
 @agate-dark-gray:   @silicon-dark-two;
 @agate-gray:        #8a8a8a;
-@agate-light-gray:  #ededed;
+@agate-light-gray:  #cccccc;
```
ybsung, line 55: How about `@agate-very-light-gray`?
https://www.htmlcsscolor.com/hex/CCCCCC

`#D0D0D0` could be `light-gray`, I think. This color could be used later. Just in case.

Popup/Popup.js
```diff
@@ -71,19 +81,21 @@ const PopupBase = kind({
 		className: 'popup'
 	},
 	computed: {
-		className: ({closeButton, title, styler, centered}) => styler.append({withCloseButton: closeButton, withTitle: title, centered})
+		className: ({centered, closeButton, onScreenDisplay, styler, title}) => styler.append({onScreenDisplay: onScreenDisplay, withCloseButton: closeButton, withTitle: title, centered})
```
ybsung, line 84: It could be `styler.append({onScreenDisplay,`.

Popup/Popup.js
```diff
@@ -71,19 +81,21 @@ const PopupBase = kind({
...
 	},
-	render: ({buttons, children, closeButton, css, noAnimation, onClose, onHide, open, skin, title, ...rest}) => {
+	render: ({buttons, children, closeButton, css, noAnimation, onClose, onHide, open, onScreenDisplay, skin, title, ...rest}) => {
```
ybsung, line 86: Please order props alphabetically.

styles/colors-silicon-day.less
daniel-stoian-lgp: good idea. i just didn't know a good name for the new value

daniel-stoian-lgp: now I see that in color-base.less we already have: @agate-light-gray:  #ccc;
Would it be a better idea for silicon to just inherit this value?

ybsung: Oh I got it. Let's inherit it first in the PR, then rename it later when needing D0D0D0 :)

Popup/Popup.js
```diff
@@ -56,6 +56,16 @@ const PopupBase = kind({
...
 		onHide: PropTypes.func,
+
+		/**
+		 * Puts the popup on top of screen.
+		 *
+		 * @type {Boolean}
+		 * @default false
+		 * @public
+		 */
+		onScreenDisplay: PropTypes.bool,
```
seunghoh, line 67: `onScreenDisplay` name is not straightforward for me. 
What about to have `position` like
```
position: PropTypes.oneOf(['bottom', 'center', 'fullscreen', 'left', 'right', 'top']),
```
for more flexibility?
As of now, we can probably have only `center` and `top` but we can support with the same prop when the new requirement comes like `bottom` `right` left` and even `fullscreen`
@ybsung FYI, I'd like to have the same prop as Popup in Sandstone

daniel-stoian-lgp:
> And one more thing, this is not related to this PR, but I also would like to change prop `centered`
> Why this should be boolean? Can't we have `left` `center` and `right`?
> This prop works for text align so I'd like to have clearer name something like `textAlign`
> Please correct me know if I am misunderstanding
> @ybsung FYI

I personally agree with this approach. Looking forward for other opinions.
We can implement this change in the near future

Popup/Popup.js
daniel-stoian-lgp: ok, i renamed it to "position". Default is center and "top" is also handled

Popup/Popup.js
```diff
@@ -57,33 +57,46 @@ const PopupBase = kind({
...
 		open: PropTypes.bool,
+
+		/**
+		 * Sets the position of the popup on the screen.
+		 *
+		 * @type {('bottom'|'center'|'fullscreen'|'left'|'right'|'top')}
+		 * @default 'center'
+		 * @public
+		 */
+		position: PropTypes.oneOf(['bottom', 'center', 'fullscreen', 'left', 'right', 'top']),
```
seunghoh, line 68: Code looks we only support `center` and `top` right? then let's remove rest

daniel-stoian-lgp: I removed the extra values

samples/sampler/stories/default/Popup.js
```diff
@@ -4,32 +4,45 @@ import {boolean, select, text} from '@enact/storybook-utils/addons/knobs';
...
+						scrimType={select('scrimType', ['none', 'translucent', 'transparent'], Config, 'translucent')}
+						spotlightRestrict={select('spotlightRestrict', ['self-first', 'self-only'], Config, 'self-only')}
+						title={text('title', Config, 'Title')}
+					>
+						<div>{text('children', Config, 'Hello Popup')}</div>
+						{buttons ? (
+							<buttons>
+								<Button>NO</Button>
+								<Button>YES</Button>
+							</buttons>
```
ybsung, line 38: `buttons` is true or undefined. It's not a component name.

Anyway, I hope that we could have Popup qa sampler with buttons because it is not a basic usage of Popup.

ybsung: Hello @daniel-stoian-lgp 
I updated sampler and qa sample for Popup according to my comment above.
But the current look is different from the GUI. Could you update it?

ybsung:
Currently agate/Popup was not wrapped with SpotlightContainer. I don't know why.
Could you define TransitionContainer like `moonstone` in https://github.com/enactjs/moonstone/blob/develop/Popup/Popup.js#L33 and use it instead of `Transition` in Popup like https://github.com/enactjs/moonstone/blob/develop/Popup/Popup.js#L225.

```
const TransitionContainer = SpotlightContainerDecorator(
	{enterTo: 'default-element', preserveId: true},
	Transition
);
```

```
		return (
			<TransitionContainer // <Transition for now in Agate develop branch
```

samples/sampler/stories/default/Popup.js
daniel-stoian-lgp: Yes, I will update. The buttons must be declared inside a <buttons> block in order to get the styling

ybsung: Oh I misunderstood. You used `buttons` as a slot not a tag name.

styles/variables-base.less
```diff
@@ -336,21 +336,28 @@
 // Popup
 // ---------------------------------------
 @agate-popup-min-width: 510px;
-@agate-popup-max-width: 70%;
```
ybsung, line 339: Why did you remove the @agate-popup-max-width variable? It will break our UX.

daniel-stoian-lgp: I saw it had no effect on the Popup on all skins

Popup/Popup.module.less
```diff
@@ -62,12 +81,18 @@
 
 		&.withCloseButton {
 			.title {
-				padding: @agate-popup-title-padding;
+				padding: @agate-popup-closebutton-title-padding;
```
ybsung, line 84: In RTL, the padding is broken.
image

Please check the comment that I left.
https://github.com/enactjs/agate/pull/276#discussion_r450063992

styles/variables-silicon.less
```diff
@@ -136,18 +136,25 @@
...
+@agate-popup-body-border-radius: 0 0 @agate-popup-border-radius @agate-popup-border-radius;
+@agate-popup-buttons-border-radius: @agate-popup-body-border-radius;
+@agate-popup-buttons-padding: @agate-component-spacing 0;
+@agate-popup-content-padding: @agate-component-spacing @agate-component-spacing (3 * @agate-component-spacing) @agate-component-spacing;
 @agate-popup-closebutton-position-left: auto;
 @agate-popup-closebutton-position-right: 0;
+@agate-popup-opacity: 0.95;
+@agate-popup-top-border-radius: 0 0 @agate-popup-border-radius @agate-popup-border-radius;
+@agate-popup-top-margin-left: 21px;
+@agate-popup-top-width: 100%;
```
ybsung, line 156: This line will be overridden by the next line.

styles/variables-base.less
ybsung: If adding `style={{width: '100%'}}` in the Popup sampler
```
storiesOf('Agate', module)
	.add(
		'Popup',
		() => (
			<div>
				<Popup
					...
					style={{width: '100%'}}
				>
```

in `develop` branch
image

in your branch
image

The popup width could be resized by app developers. The GUI developers wanted not to expand the popup width to the same size with the screen width. So that's the reason why we added the max-width.

ybsung:
Hello @daniel-stoian-lgp 

I think your PR is almost ready. Could you check my comments?

styles/variables-base.less
daniel-stoian-lgp: ok, I reverted the change :)

Popup/Popup.module.less
daniel-stoian-lgp: ok, I will fix

styles/variables-base.less
daniel-stoian-lgp: reverted this in variables-base but kept the removal in variables-gallium and variables-silicon as they had the same value

Popup/Popup.js
```diff
@@ -57,33 +64,46 @@ const PopupBase = kind({
...
 	},
 	computed: {
-		className: ({closeButton, title, styler, centered}) => styler.append({withCloseButton: closeButton, withTitle: title, centered})
+		className: ({centered, closeButton, position, styler, title}) => styler.append({top: position === 'top', withCloseButton: closeButton, withTitle: title, centered})
 	},
-	render: ({buttons, children, closeButton, css, noAnimation, onClose, onHide, open, skin, title, ...rest}) => {
+	render: ({buttons, children, closeButton, css, noAnimation, onClose, onHide, open, position, skin, title, ...rest}) => {
 		const wideLayout = (skin === 'carbon');
+		const transitionType = (position === 'center' ? 'fade' : 'slide');
+		const direction = (position === 'center' ? 'down' : 'up');
```
Djspaceg, line 97: These seem like great candidates for `computed` props.

Popup/Popup.module.less
```diff
@@ -40,6 +48,19 @@
 		border-width: @agate-popup-border-width;
 		border-color: @agate-popup-border-color;
 		border-radius: @agate-popup-border-radius;
+		opacity: @agate-popup-opacity;
+
+		&.top {
+			border-radius: @agate-popup-top-border-radius;
+			margin: @agate-popup-top-margin;
+			max-width: @agate-popup-top-max-width;
+			width: @agate-popup-top-width;
```
Djspaceg, line 57: Rather than setting these sorts of sizing things in variables, let's make the `.top` class a pre-defined shape. The top and left rules aren't variable, so this would only be able to stick to the top left corner and be various widths from there. That seems less useful. However, if we instead specify `right` up above, on line 25, we don't need any of the width props here, and the content can size itself.

If you wanted to be _really fancy_ (and who wouldn't‽), you could set it to anchor in the top center of the screen, and allow the content to dictate the width, up to 100% screen width, so a skin could choose its width and it always stay centered. This is more of a bonus, and not mandatory.

Popup/Popup.module.less
```diff
@@ -62,12 +83,20 @@
 
 		&.withCloseButton {
 			.title {
-				padding: @agate-popup-title-padding;
```
Djspaceg, line 65: I'd like to avoid this practice, that is: removing a generalized shorthand rule and switching it out for several longhand rules, and abandoning the single shorthand multi-variable for many specifically named variables. It makes the variables inconsistent and harder to predict and maintain.

LESS offers an `extract` function that can pull values out of shorthand representations in variables. The problem with it is not very forgiving if you extract the 4th argument and there are only 2, for example. For this reason, I've created a mixin that works like the function, but safely handles all lengths of shorthand values. _Code at the bottom of this comment._ Eventually, we'll get this mixin into the core enact package, so we don't need it in the themes, but for now, this is the way it must be. We can take it out once it's in the core.

### Usage

Currently, it is primarily intended for shorthand 4-value rules, like: margin, padding, .position(), and even border-(width|style|color), border-radius, etc.

Standalone, to grab just the left shorthand value:
```.extract(@agate-popup-title-padding, left)[]```
_Important to remember the trailing `[]` so you get a variable back._

In this case:
```less
padding: @agate-popup-title-padding;
.padding-start-end(.extract(@agate-popup-title-padding, left)[], .extract(@agate-popup-title-padding, right)[]);
```

### Code

The following should be placed in a file here: `styles/internal/utils.less`, and imported into both `variables-base.less` and `mixins.less`.

`@import "./internal/utils.less";`

```less

//
// Computational mixins that return values
//


// Normalizes a list of values, using the logic from padding and margin short-hand, and expands it up
// to the full long-hand representation, giving a consistent list length to extract from.
// Supports 1, 2, 3, or 4 values in a space separated list.
// Always returns a list of 4.
// .normalize-shorthand(11px 21px 31px 41px)[]  -->  11px 21px 31px 41px
// .normalize-shorthand(11px 21px 31px)[]       -->  11px 21px 31px 21px
// .normalize-shorthand(11px 21px)[]            -->  11px 21px 11px 21px
// .normalize-shorthand(11px)[]                 -->  11px 11px 11px 11px
//
// NOTE: The [] is required at the end of the method call, which instructs LESS to return a value.
// For details see: http://lesscss.org/features/#mixins-feature-unnamed-lookups
.normalize-shorthand(@trbl) when ((length(@trbl) >= 1) and (length(@trbl) <= 4)) {
	@result:
		extract(@trbl, 1)
		extract(@trbl, if((length(@trbl) >= 2), 2, 1) )
		extract(@trbl, if((length(@trbl) >= 3), 3, 1) )
		extract(@trbl, if((length(@trbl) >= 4), 4, if((length(@trbl) >= 2), 2, 1) ))
	;
}

// Extracts one or all short-hand values out of a list (See `.normalize-shorthand`), returning all
// values or a requested value.
// padding-left: .extract(11px 21px 31px 41px, left)[]  -->  41px
// padding-left: .extract(11px 21px 31px, left)[]       -->  21px
// padding-left: .extract(11px 21px, left)[]            -->  21px
// padding-left: .extract(11px, left)[]                 -->  11px
// padding: .extract(11px 21px 31px 41px)[]             -->  11px 21px 31px 41px
//
// NOTE: The [] is required at the end of the method call, which instructs LESS to return a value.
// For details see: http://lesscss.org/features/#mixins-feature-unnamed-lookups
.extract(@trbl; @position) when (@position = top) {
	@result: extract(.normalize-shorthand(@trbl)[], 1);
}
.extract(@trbl; @position) when (@position = right) {
	@result: extract(.normalize-shorthand(@trbl)[], 2);
}
.extract(@trbl; @position) when (@position = bottom) {
	@result: extract(.normalize-shorthand(@trbl)[], 3);
}
.extract(@trbl; @position) when (@position = left) {
	@result: extract(.normalize-shorthand(@trbl)[], 4);
}
.extract(@trbl) {
	@result: .normalize-shorthand(@trbl)[];
}

// Extracts one value out of a short-hand list, given a position. Works like the built-in `extract`
// LESS function, but is safe to use on lists of arbitrary or unknown length.
// Always returns 1 value.
// padding-left: .extract(11px 21px 31px 41px, 4)[]  -->  41px
// padding-left: .extract(11px 21px 31px, 4)[]       -->  21px
// padding-left: .extract(11px 21px, 4)[]            -->  21px
// padding-left: .extract(11px, 4)[]                 -->  11px
//
// NOTE: The [] is required at the end of the method call, which instructs LESS to return a value.
// For details see: http://lesscss.org/features/#mixins-feature-unnamed-lookups
.extract(@trbl; @position) when ((@position >= 1) and (@position <= 4)) {
	@result: extract(.normalize-shorthand(@trbl)[], @position);
}

```

styles/colors-silicon-day.less
```diff
@@ -52,7 +52,6 @@
 @agate-darker-gray: @silicon-dark;
 @agate-dark-gray:   @silicon-dark-two;
 @agate-gray:        #8a8a8a;
-@agate-light-gray:  #ededed;
```
Djspaceg, line 55: Was this intentional?

styles/variables-gallium.less
```diff
@@ -104,12 +104,12 @@
...
 // ---------------------------------------
 @agate-popup-min-width: 510px;
-@agate-popup-max-width: 70%;
 @agate-popup-border-width: 3px;
 @agate-popup-border-radius: 20px;
 @agate-popup-border: 3px solid transparent;
 @agate-popup-padding: (@agate-component-spacing * 2.5) (@agate-component-spacing * 5);
-@agate-popup-title-padding: 0 (@agate-button-small-height) 0 0;
+@agate-popup-closebutton-title-padding-left: 0;
+@agate-popup-closebutton-title-padding-right: @agate-button-small-height;
```
Djspaceg, line 112: I think this type of change makes assumptions that each skin will have a close button that borrows space from the title. Is there a different approach we could use that leverages the existing variables?

styles/variables-silicon.less
```diff
@@ -152,18 +152,25 @@
...
+@agate-popup-title-padding: 0 @agate-component-spacing;
+@agate-popup-body-padding: (@agate-component-spacing * 2) @agate-component-spacing 0 @agate-component-spacing;
+@agate-popup-body-margin: 0;  // Extend to the edge
+@agate-popup-body-border-radius: 0 0 @agate-popup-border-radius @agate-popup-border-radius;
+@agate-popup-buttons-border-radius: @agate-popup-body-border-radius;
+@agate-popup-buttons-padding: @agate-component-spacing 0;
+@agate-popup-content-padding: @agate-component-spacing @agate-component-spacing (3 * @agate-component-spacing) @agate-component-spacing;
 @agate-popup-closebutton-position-left: auto;
 @agate-popup-closebutton-position-right: 0;
+@agate-popup-opacity: 0.95;
```
Djspaceg, line 170: The entire popup is 95% opacity? even the content? What if we move this to just the background color? `fade(@bgcolor, 95%)`

styles/variables-gallium.less
daniel-stoian-lgp: I don't know. All skins used this padding so far when a close button was present. The only thing I did was to break this padding into padding-left and padding-right to support RTL

PR #280, opened 2020-06-29
FORTECH-61 Update ImageItem with text over image

adrian-cocoara-lgp:
Added 'captionPosition' prop to select the position of caption.


ImageItem/ImageItem.js
```diff
@@ -43,6 +43,14 @@ const ImageItemBase = kind({
 		 */
 		children: PropTypes.node,
 
+		/**
+		 * Display caption over image.
+		 *
+		 * @type {String}
```
daniel-stoian-lgp, line 49: type {Boolean}

ImageItem/ImageItem.js
```diff
@@ -43,6 +43,14 @@ const ImageItemBase = kind({
...
+		 * @public
+		 */
+		childrenOverImage: PropTypes.bool,
```
daniel-stoian-lgp, line 52: maybe it is better to rename the prop "captionOverImage" ? I saw you used  this also as a class name

ImageItem/ImageItem.js
```diff
@@ -88,20 +96,32 @@ const ImageItemBase = kind({
...
 
-	render: ({children, css, src, ...rest}) => {
-		const [Component, marqueeProps] = children ? [MarqueeImageItem, {
+	computed: {
+		className: ({childrenOverImage, styler}) => styler.append({
+			captionOverImage: childrenOverImage
+		})
+	},
+
+	render: ({children, css, childrenOverImage,  src, ...rest}) => {
```
daniel-stoian-lgp, line 105: please sort the props alphabetically

ImageItem/ImageItem.module.less
```diff
@@ -11,7 +11,15 @@
 		}
   	}
 
-  	.applySkins({
+	&.captionOverImage {
+		.caption {
+			overflow: hidden;
+			position: absolute;
+			bottom: 6px;
```
daniel-stoian-lgp, line 18: maybe this could be a variable, only in variables-base. it will be easier later to customize this for other skins

ybsung:
Please update the description for `childrenOverImage` which might be `captionOverImage`.

ImageItem/ImageItem.js
```diff
@@ -35,6 +35,14 @@ const ImageItemBase = kind({
 	name: 'ImageItem',
 
 	propTypes: /** @lends agate/ImageItem.ImageItemBase.prototype */ {
+		/**
+		 * Display caption over image.
+		 *
+		 * @type {Boolean}
+		 * @public
+		 */
+		captionOverImage: PropTypes.bool,
```
ybsung, line 44: As @seunghoh mentioned in https://github.com/enactjs/agate/pull/279#discussion_r447323032, how about providing a `type` prop including `fullimage` or something else instead of providing a boolean type prop?

ImageItem/ImageItem.js
```diff
@@ -88,20 +96,32 @@ const ImageItemBase = kind({
...
+				<Component
+					{...rest}
+					{...marqueeProps}
+					css={css}
+					imageComponent={ImageBase}
+					src={src}
+				>
+					{children}
+				</Component>
+			</React.Fragment>
```
ybsung, line 124: Do we need the `React.Fragment`?

ImageItem/ImageItem.js
```diff
@@ -88,20 +96,32 @@ const ImageItemBase = kind({
...
-	render: ({children, css, src, ...rest}) => {
-		const [Component, marqueeProps] = children ? [MarqueeImageItem, {
+	computed: {
+		className: ({captionOverImage, styler}) => styler.append({
+			captionOverImage
+		})
+	},
+
+	render: ({captionOverImage, children, css,  src, ...rest}) => {
+		const imageItemWithOverlayCaption = (children && captionOverImage) ? [UiImageItem, children] : [UiImageItem, null];
```
ybsung, line 106: If children && captionOverImage, then `children` is passed to `Component` twice. 
One for `{...marqueeProps}`, the other for `{children}`.

I think that the conditions get complex. Let's make the code more comprehensible.

ybsung: I couldn't understand why not `[UiImageItem, null]` but `[UiImageItem, children]`?

adrian-cocoara-lgp: You are right, this works if I use [UiImageItem, null].

ImageItem/ImageItem.js
adrian-cocoara-lgp: No we do not. I will delete it.

ImageItem/ImageItem.js
```diff
@@ -88,20 +96,32 @@ const ImageItemBase = kind({
...
+
+		const [Component, marqueeProps] = (children && !captionOverImage) ? [MarqueeImageItem, {
 			alignment: 'center',
 			children
-		}] : [UiImageItem, null];
+		}] : imageItemWithOverlayCaption;
```
ybsung, line 111: Can we replace your code with the following code?
```
		const [Component, marqueeProps] = children ? 
			[MarqueeImageItem, {alignment: 'center'}] :
			[UiImageItem, null];
```
In my opinion, it would be more understandable. Please let me know your thought.

ImageItem/ImageItem.module.less
```diff
@@ -11,7 +11,14 @@
 		}
   	}
 
-  	.applySkins({
+	&.captionOverImage {
+		.caption {
+			overflow: hidden;
```
ybsung, line 16: What is our UX/GUI for too long texts? Clipping it, adding `...`, or something else?

ImageItem/ImageItem.js
adrian-cocoara-lgp: This code would apply Marquee to caption when placed on image, but we don't want this to happen. This code works:
`const [Component, marqueeProps] = (children && !captionOverImage) ? 
[MarqueeImageItem, {
	alignment: 'center',
	children
}] : [UiImageItem, null];`

ybsung:
Hello @adrian-cocoara-lgp 
I hope I could review the less files later. Please check my comments above first?

ImageItem/ImageItem.js
ybsung: Ooops. I added the wrong condition. The condition that you wrote is the one that I expected.
By the way, do we have to pass the `children` in `{ alignment: 'center', children }` because there is the duplicated `children` in `<Component ...{ alignment: 'center', children } ... >{children}</Component>`?

ImageItem/ImageItem.js
adrian-cocoara-lgp: I changed `captionOverImage: PropTypes.bool` with `captionPosition: PropTypes.oneOf(['captionBelow', 'captionOverlay'])` with default: `captionBelow`.
What is your opinion on this approach?

ImageItem/ImageItem.js
adrian-cocoara-lgp: You're right. We don't need to pass `children` in `{ alignment: 'center', children }`. I will remove it.

ImageItem/ImageItem.module.less
adrian-cocoara-lgp: For now we don't have UX/GUI design when text is long. I went with this approach, but it can be changed if required.

ImageItem/ImageItem.js
ybsung: Much better~
Only one thing, I'd like to comment..
How about replacing caption with children?
App developers couldn't know the relationship between caption and children.

ImageItem/ImageItem.js
ybsung: Good to know~

ImageItem/ImageItem.module.less
ybsung: I got it. Let's support it separately if needed later.

ImageItem/ImageItem.js
```diff
@@ -43,6 +43,16 @@ const ImageItemBase = kind({
...
 
+		/**
+		 * Sets the position for caption.
+		 * Available positions: 'captionBelow' (default) and 'captionOverlay'.
+		 *
+		 * @type {('captionBelow'|'captionOverlay')}
+		 * @default 'captionBelow'
+		 * @public
+		 */
+		childrenPosition: PropTypes.oneOf(['captionBelow', 'captionOverlay']),
```
ybsung, line 54: @adrian-cocoara-lgp 
We're still thinking what is the best name and values for this prop.
We'll let you know after discussed.

Considering the following names
- childrenPosition
- captionBelow & captionOverlay
- captionOverImage

ybsung: Our decision is..
childrenPosition -> captionPosition // `caption` is more clear to app developers
captionBelow -> below // `caption` is already included in the prop name - `captionPosition`
captionOverlay -> overlay // same above
captionOverImage -> captionOverlay // Using just `overlay` could make confused

adrian-cocoara-lgp: `captionOverImage` was replaced with `childrenPosition`. I don't see any reference to it anywhere, do we still need it?

ybsung: > captionOverImage was replaced with childrenPosition. 

captionOverImage was replaced with captionOverlay
 
> Do you mean the captionOverImage className from less file?

Yes

So the className in computed would be..
```
		className: ({captionPosition, styler}) => styler.append({
			captionOverlay: captionPosition === 'overlay'
		})
```

PR #281, opened 2020-06-30
FORTECH-60: Item with subtext and icons

stanca-pop-lgp:
Feature Added
Updated styles for Agate Item component.

Comments
Updated Agate Item component styling according to latest Zeplin design. Added stories for "List Item with Small Icon" (SmallIconsItem) and "Multi Selection" (IconSubtextItem) in Agate Storybook. Added unit test.
Removed variable _@agate-item-selected-outline: 5px solid @agate-item-focus-bg-color;_ as it was not used.


styles/variables-silicon.less
```diff
@@ -104,13 +112,14 @@
...
 // ---------------------------------------
-@agate-item-font-size: 40px;
-@agate-item-font-weight: 500;
-@agate-item-label-font-size: 30px;
-@agate-item-side-outset: 32px;
-@agate-item-height: 90px;
-@agate-item-height-large: 120px;
-@agate-item-border-width: 3px;
+@agate-item-border-width:     3px;
+@agate-item-font-size:        1.2rem;
```
daniel-stoian-lgp, line 116: why did you switch from px to rem?

samples/sampler/stories/default/SmallIconsItem.js
```diff
@@ -0,0 +1,40 @@
...
+						<Icon size="small">{slotAfterIcon1}</Icon>
+						<Icon size="small">{slotAfterIcon2}</Icon>
+					</slotAfter>
+				</Item>
+			);
+		},
+		{
+			text: 'Basic usage of SmallIconsItem'
+		}
+	);
```
ybsung, line 40: There is no `SmallIconsItem` public component that we provide. So we won't add `SmallIconsItem` sampler. The name sampler should be the name which is the same with one of our public components.

If not, app developers could confuse whether there is the `SmallIconsItem` public component or not. In addition, we could confuse where is the sampler to verify the Item with small icons.

@0x64 
I would like to unify the `Item` sampler, the `SmallIconsItem` sampler, and the `IconSubTextItem` sampler into the `Item` sampler. Please let me know your thoughts with the following code.
```
		() => {
			const slotBeforeIcon = select('slotBefore Icon', iconList, Config);
			const slotAfter = [
				select('slotAfter Icon 1', iconList, Config),
				select('slotAfter Icon 2', iconList, Config),
				text('slotAfter Text', Config, 'Sub Text')
			].filter(text=>!!text).map(text=>{(iconList.indexOf(text) > 0 ? <Icon key={text}>{text}</Icon> : text));

			return (
				<Item
					inline={boolean('inline', Config)}
					label={text('label', Config)}
					labelPosition={select('labelPosition', ['above', 'after', 'before', 'below'], Config, 'below')}
					selected={boolean('selected', Config)}
				>
					{slotBeforeIcon ? <slotBefore>
						<Icon>{slotBeforeIcon}</Icon>
					</slotBefore> : null}
					{text('children', Config, 'Hello Item')}
					{slotAfter.length > 0 ? <slotAfter>
						{slotAfter}
					</slotAfter> : null}
				</Item>
			);
		},
```

ybsung: After discussing with @0x64 

1. Replace SmallIconsItems.js and IconSubTextItem.js with the following code
2. Add the following code at the end of the samples/sampler/stories/default/Item.js.

```
storiesOf('Agate QA.Item', module)
	.add(
		'with icons and a text in slotAfter',
		() => {
			const slotBeforeIcon = select('slotBefore Icon', iconList, Config);
			const slotAfter = [
				select('slotAfter Icon 1', iconList, Config),
				select('slotAfter Icon 2', iconList, Config),
				text('slotAfter Text', Config, 'Sub Text')
			].filter(text=>!!text).map(text=>(iconList.indexOf(text) > 0 ? <Icon key={text}>{text}</Icon> : text));

			return (
				<Item
					inline={boolean('inline', Config)}
					label={text('label', Config)}
					labelPosition={select('labelPosition', ['above', 'after', 'before', 'below'], Config, 'below')}
					selected={boolean('selected', Config)}
				>
					{slotBeforeIcon ? <slotBefore>
						<Icon>{slotBeforeIcon}</Icon>
					</slotBefore> : null}
					{text('children', Config, 'Hello Item')}
					{slotAfter.length > 0 ? <slotAfter>
						{slotAfter}
					</slotAfter> : null}
				</Item>
			);
		}
	);

```

Item/Item.module.less
```diff
@@ -41,6 +41,7 @@
 		border-style: solid;
 		border-width: @agate-item-border-width;
 		border-color: @agate-item-border-color;
+		border-bottom-color: @agate-item-border-bottom-color;
```
ybsung, line 44: I don't think that the last item also has the border bottom
image

If the item is the last child, then the item's borer shouldn't show.
```
&:last-child {
	border-bottom-color: transparent; // or `border-bottom-width: 0` and `box-sizing: border-box` with proper item size
}
```
image

samples/sampler/stories/default/SmallIconsItem.js
0x64: Thank you @ybsung for an update from our discussion.

To be clear, here is my thought on this.

They are not components but examples that have not a default behavior of a component but have a given behavior for specific use cases. And, if changing children inside `slotAfter` is not a big issue to be provided as a sampler, I prefer to make children as a combination of icon(s) text element(s) inside `slotAfter` as a usage guide.

If we need to make other cases for testing of specific given requirements for our target domain beyond the basic behavior of the component, it would be better to add a sample into qa sampler, not default sampler.

styles/variables-silicon.less
```diff
@@ -104,13 +112,14 @@
...
 // ---------------------------------------
-@agate-item-font-size: 40px;
-@agate-item-font-weight: 500;
-@agate-item-label-font-size: 30px;
-@agate-item-side-outset: 32px;
-@agate-item-height: 90px;
-@agate-item-height-large: 120px;
-@agate-item-border-width: 3px;
+@agate-item-border-width:     3px;
+@agate-item-font-size:        28px;
```
ybsung, line 116: Let's have a multiple of 3. Maybe 27px?

ybsung:
No JSDoc parse error except the following known issue
: Parse error: unknown tag @agate in raw/agate/ThumbnailItem:6

Item/Item.module.less
seunghoh: YB is correct. Please refer the GUI feedback on this
>The last list item on the left tab menu has no underline. It's a hint for the last item. However when it selected, red line is displayed like other list items.

Item/Item.module.less
```diff
@@ -104,5 +110,24 @@
 				}
 			}
 		});
+
+		.active({
```
seunghoh, line 114: hm.. do you have any reason(or requirement) why added `active` on Item component? Enact Item actually doesn't have any pressed effect on click or touch. From my UX/GUI understanding, red dot + red underline should be applied to `selected`
@Djspaceg What do you think?

samples/sampler/stories/default/SmallIconsItem.js
stanca-pop-lgp: I understand, thank you for your help and clarifications.

Item/Item.module.less
stanca-pop-lgp: You are right. I applied it to selected. Thank you!

styles/variables-silicon.less
```diff
@@ -89,6 +89,14 @@
 // ---------------------------------------
 @agate-imageitem-border-width: 6px;
 
+// Indicator
+// ---------------------------------------
+@agate-indicator-align-self:    center;
+@agate-indicator-border-radius: @agate-indicator-height / 2;
+@agate-indicator-display:       flex;
```
Djspaceg, line 96: I don't think this is a good practice. This has layout structural implications and should be coded into the component file itself. `display:none` removes the component from the layout, rather than just hiding it like `visibility:hidden` or `opacity:0`. Let's find another approach that has a similar visual effect, but doesn't do this.

PR #282, opened 2020-07-01
FORTECH-63: Update Header with title text

adrian-cocoara-lgp:
Updated styles for 'silicon' skin
Added unit tests


styles/colors-silicon-night.less
```diff
@@ -83,6 +83,11 @@
 @agate-fullscreenpopup-bg-color: @agate-popup-bg-color;
 @agate-fullscreenpopup-bg-image: @agate-bg-image;
 
+// Header
+// ---------------------------------------
+@agate-header-subtitle-text-color: @agate-white;
+@agate-header-subtitle-opacity: 0.4;
```
daniel-stoian-lgp, line 89: duplicate variable and value

Header/tests/Header-specs.js
```diff
@@ -0,0 +1,45 @@
...
+				titleAbove="title above"
+			/>
+		);
+
+		const expected = 'title above';
+		const actual = subject.prop('titleAbove');
+
+		expect(actual).toBe(expected);
+	});
+});
```
ybsung, line 45: Thanks for more unit tests.

styles/colors-silicon-day.less
```diff
@@ -130,6 +130,11 @@
 @agate-fullscreenpopup-bg-color: @agate-popup-bg-color;
 @agate-fullscreenpopup-bg-image: @agate-bg-image;
 
+// Header
+// ---------------------------------------
+@agate-header-subtitle-text-color: @agate-black;
```
Djspaceg, line 135: How about we set this to be `@agate-foreground`, and then delete the night rules entirely? That way, when we redefine night's `@agate-foreground` to be white, the header automatically picks up that color?

Header/Header.module.less
```diff
@@ -46,6 +46,16 @@
...
 		}
 
+		.subtitle {
+			color: @agate-header-subtitle-text-color;
+			opacity: @agate-header-subtitle-opacity;
+			padding-top: @agate-header-subtitle-padding;
+		}
+
+		.titleAbove {
+			padding-bottom: @agate-header-subtitle-padding;
```
Djspaceg, line 56: I think this should be broken out into its own variable, specifically for `titleabove`.

Header/Header.module.less
```diff
@@ -46,6 +46,16 @@
 			text-indent: @agate-header-title-text-indent;  // Fix for characters like `j` being left-truncated because of being too close to the edge
 		}
 
+		.subtitle {
+			color: @agate-header-subtitle-text-color;
+			opacity: @agate-header-subtitle-opacity;
+			padding-top: @agate-header-subtitle-padding;
```
Djspaceg, line 52: This appears to work now only because the padding value is set to a single value, not a shorthand multi-value. This rule should switch to just plain `padding` and the `@agate-header-subtitle-padding` variable should become `9px 0 0 0`.

styles/variables-base.less
```diff
@@ -183,7 +183,8 @@
 @agate-header-title-margin: 0;
 @agate-header-title-font-size: inherit;
 @agate-header-title-text-indent: inherit;
-@agate-header-subtitle-font-size: 27px;
+@agate-header-subtitle-font-size: 24px;
+@agate-header-subtitle-padding: 9px;
```
Djspaceg, line 187: The padding here in the base variables file (which applies to every skin) should be set to 0, then set to 9px in silicon and gallium variables files.

styles/colors-silicon-day.less
adrian-cocoara-lgp: Seems like a good idea!

PR #283, opened 2020-07-07
FORTECH-67: Silicon styles for Dropdown

daniel-stoian-lgp:
**Feature Added**
Added silicon styles for Dropdown


Dropdown/Dropdown.js
```diff
@@ -184,37 +196,77 @@ const DropdownBase = kind({
...
+								>
+									{children || []}
+								</Group>
+							</Scroller>
+						</ContainerDiv>
+					</Transition>
+				</div>
+				}
+				{!dropdownButton &&
+					<React.Fragment>
```
adrian-cocoara-lgp, line 239: Do we need React.Fragment here?

daniel-stoian-lgp: yes, we need an enclosing tag, But I did not want to add an unnecessary<div>

Dropdown/Dropdown.js
```diff
@@ -184,37 +196,77 @@ const DropdownBase = kind({
...
+						className={transitionContainerClassname}
+						visible={opened}
+						direction={transitionDirection}
+					>
+						<ContainerDiv className={dropdownListClassname} spotlightDisabled={!open} spotlightRestrict="self-only">
+							<Scroller className={css.scroller}>
+								<Group
+									className={css.group}
+									childComponent={RadioItem}
+									itemProps={{size: 'small', className: css.dropDownListItem, css: css}}
```
ybsung, line 226: We could use `, css}}` simply instead of `, css: css}}`.

Dropdown/Dropdown.js
```diff
@@ -184,37 +196,77 @@ const DropdownBase = kind({
...
+										onSelect={onSelect}
+										selected={selected}
+									>
+										{children || []}
+									</Group>
+								</Scroller>
+							</ContainerDiv>
+						</Transition>
+					</React.Fragment>
+				}
```
ybsung, line 269: I don't like the two pair of the components depending on the `silicon` skin.
Please check my suggestion and let me know your thoughts.

The commit based on your PR: https://github.com/enactjs/agate/commit/cf3829c4f9207055c71ea072d23bf087f8fea909

```suggestion
const [DropDownButton, wrapperProps, groupProps] = (skin === 'silicon') ? [
	Button,
	{className: dropdownButtonClassname, component: 'div'},
	{childComponent: RadioItem, className: css.dropDownListItem, css, itemProps: {size: 'small', selectedProp: 'selected'}}
] : [
	Item,
	{component: React.Fragment},
	{childComponent: Item, itemProps: {size: 'small'}}
];

return (
	<div {...rest}>
		<ComponentOverride {...wrapperProps}>
			<DropDownButton
				{...rest}
				css={css}
				disabled={hasChildren ? disabled : true}
				onClick={opened ? onClose : onOpen}
			>
				<Icon slot="slotAfter" className={css.icon} size="small">{open ? 'arrowlargeup' : 'arrowlargedown'}</Icon>
				{title}
			</DropDownButton>
			<Transition
				className={transitionContainerClassname}
				visible={opened}
				direction={transitionDirection}
			>
				<ContainerDiv className={dropdownListClassname} spotlightDisabled={!open} spotlightRestrict="self-only">
					<Scroller className={css.scroller}>
						<Group
							className={css.group}
							onSelect={onSelect}
							selected={selected}
							{...groupProps}
						>
							{children || []}
						</Group>
					</Scroller>
				</ContainerDiv>
			</Transition>
		</ComponentOverride>
	</div>
);
```

daniel-stoian-lgp: applied the suggested change with minor adjustments. see  story ticket

samples/sampler/stories/default/Dropdown.js
```diff
@@ -33,3 +32,56 @@ storiesOf('Agate', module)
...
+storiesOf('Agate QA.Dropdown', module)
+	.add(
+		'Dropdown directions',
+		() => {
+			const itemCount = number('items', Config, {range: true, min: 0, max: 50}, 5);
+			const items = (new Array(itemCount)).fill().map((i, index) => `Option ${index + 1}`);
+
+			return (
+				<div>
+					<div style={{height: '350px'}}>
```
ybsung, line 45: Please do not use pixel unit in samplers. I recommend you to use ri.scaleToRem() instead of `px`.

daniel-stoian-lgp: done

Dropdown/Dropdown.js
```diff
@@ -156,7 +166,9 @@ const DropdownBase = kind({
 	},
 
 	computed: {
+		className: ({open, styler}) => styler.append({listOpen: open, listClosed: !open}),
```
ybsung, line 169: `listClosed` might not used anywhere. Could we remove it?
And in my opinion, it would be just fine to use `open` instead of `listOpen`.

samples/sampler/stories/default/Dropdown.js
```diff
@@ -33,3 +33,52 @@ storiesOf('Agate', module)
...
+						onSelect={action('onSelect')}
+						style={{position: 'absolute', top: 0, left: '50%', transform: 'translateX(-50%)', width:'25%'}}
+						title="Down dropdown"
+					>
+						{items}
+					</Dropdown>
+					<Dropdown
+						direction="up"
+						onSelect={action('onSelect')}
+						style={{position: 'absolute', top: ri.scaleToRem(350), width:'25%'}}
```
ybsung, line 73: Could we change `350` to `440` to show the dropdown properly in a viewport because it height is 435px?

Dropdown/Dropdown.js
```diff
@@ -184,37 +196,49 @@ const DropdownBase = kind({
...
-								onSelect={onSelect}
-								selected={selected}
-							>{children || []}</Group>
-						</Scroller>
-					</ContainerDiv>
-				</Transition>
+			<div className={className} {...rest} >
+				<div {...wrapperProps}>
+					<DropDownButton
+						className={className}
```
ybsung, line 215: Good to know that you found the solution to fix the dropdown position.

But I don't think that we need to add all the same CSS classes in the `className` to the `DropDownButton`.
For now `.dropdown`, `.silicon`, `.listOpen`, and `.listClosed` are passed to the button.
1) `listClosed` will be removed.
2) `dropdown` and `silicon` might not be passed to the button.
3) So let's pass only the `.listOpen` to it after renaming it as `open` and the `className` as `buttonClassName`.

```
	computed: {
		buttonClassName: ({open, styler}) => styler.append({open}),
		...
	render: ({buttonClassName, children, css, dropdownButtonClassname, dropdownListClassname, disabled, hasChildren, onClose, onOpen, onSelect, open, selected, skin, transitionContainerClassname, transitionDirection, title, ...rest}) => {
		...

		return (
			<div {...rest} >
				<div {...wrapperProps}>
					<DropDownButton
						className={buttonClassName}
						css={css}
						disabled={hasChildren ? disabled : true}
						onClick={opened ? onClose : onOpen}
					>
						...
					</DropDownButton>
```

samples/sampler/stories/default/Dropdown.js
ybsung: image


I found there was one issue.
After expanding "Up dropdown", if clicking the "Right dropdown" to be expanded,
the "Right dropdown" expands under the "Up dropdown". So the menus in the "Right dropdown" can't be seen.

ybsung: I created the FORTECH-83 issue for the issue above. Let's solve it separately.

Djspaceg:
Nice.

Now, technically, when altering the `skinVariants` prop, you should not only set the value, but read in the prop, spread the object onto your new prop, and then make any alterations. This avoids the problem of clobbering an existing variant setting that your component doesn't care about, but something else may. That being said, this is a corner case, so we don't need to address it now.
See: https://enactjs.com/docs/modules/ui/Skinnable/#Skinnable.defaultConfig.variantsProp
And: https://enactjs.com/docs/modules/ui/Skinnable/#Skinnable.skinVariants

PR #291, opened 2020-07-13
FORTECH-78: Removed duplicate gallium variables

adrian-cocoara-lgp:
Removed duplicate gallium variables

### DIFF
--- a/styles/colors-gallium-night.less
+++ b/styles/colors-gallium-night.less
@@ -21,7 +21,7 @@
 
 // Button
 // ---------------------------------------
-@agate-button-focus-color:    @agate-spotlight-color;
+@agate-button-color:          @agate-foreground;
 @agate-button-lightOpaque-color: @agate-black;
 @agate-button-lightOpaque-bg-color: @agate-light-gray;
 
--- a/styles/colors-gallium-night.less
+++ b/styles/colors-gallium-night.less
@@ -3,178 +3,56 @@
 
 
 // Named Colors
-@agate-accent: var(--agate-accent-color, #8b7efe);
-@agate-accent-h: var(--agate-accent-h, 246);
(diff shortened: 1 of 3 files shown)


styles/colors-gallium-night.less
```diff
@@ -3,178 +3,56 @@
...
 // ---------------------------------------
 @agate-spotlight-color:                @agate-foreground-focus;
 @agate-spotlight-bg-color:             @agate-background-focus;
-@agate-spotlight-border-color:         transparent;
 @agate-spotlight-disabled-text-color:  fade(@agate-spotlight-color, 40%);
-@agate-spotlight-focus-bg-image:       none;
 
 // Button
 // ---------------------------------------
-@agate-button-color:          @agate-foreground;
```
daniel-stoian-lgp, line 50: not the same value

styles/colors-gallium-night.less
```diff
@@ -3,178 +3,56 @@
...
-@agate-button-bg-color:       @agate-dark-gray;
-@agate-button-bg-image:       none;
-@agate-button-border-color:   transparent;
-@agate-button-shadow:         none;
 @agate-button-focus-color:    @agate-spotlight-color;
```
daniel-stoian-lgp, line 24: missed this one

ybsung:
I refreshed the branch because of the conflict.

Djspaceg:
I brought back just the two deleted core variables. It seemed wrong to not have them there, though there's no technical reason they must exist, since they're already defined as those colors. These variables at the top are a little special, so I wanted to keep discoverability of them high, rather than only existing in the base files like the other component-specific variables.

Also, I switched over the keyword colors to their corresponding agate defined color variables as a good practice. "black" doesn't match "@agate-black" so it was not switched.

PR #295, opened 2020-07-20
PLAT-112159: Support `default skin styles` knob to use default accent and highlight colors

ybsung:
Even though changing the `skin`, the `accent` and the `highlight` colors were preserved. So it was hard to verify the updated CSS styles based on the default `accent` and the `highlight` colors in the changed skin.

The `default skin styles` knob is now supported to use default `accent` and `highlight` colors instead of the preserved colors. If checking it, then the `accent` and `highlight` color knobs will be hidden.

Step to reproduce:
1. Select `Agate > Button`
2. Select `Global Knobs`
3. Select `Cobalt` as the `skin`
4. Check `default skin styles`

Expected Result: 
Button background color changes from  #E16253 to white because the highlight color of the Cobalt is white.

Before PR:
image

After PR:
image

### DIFF
--- a/samples/sampler/src/ThemeEnvironment/ThemeEnvironment.js
+++ b/samples/sampler/src/ThemeEnvironment/ThemeEnvironment.js
@@ -196,11 +196,11 @@ const StorybookDecorator = (story, config) => {
 			locale={locale}
 			{...skinKnobs}
 			skinVariants={boolean('night mode', Config) && 'night'}
-			accent={useSkinDefaultStyles ?
+			accent={useSkinDefaultStyles && !allSkins ?
 				defaultColors[skinKnobs.skin].accent :
 				color('accent', (!newSkin && accentFromURL ? accentFromURL : defaultColors[currentSkin].accent), Config.groupId)
 			}
-			highlight={useSkinDefaultStyles ?
+			highlight={useSkinDefaultStyles && !allSkins ?
 				defaultColors[skinKnobs.skin].highlight :
 				color('highlight', (!newSkin && highlightFromURL ? highlightFromURL : defaultColors[currentSkin].highlight), Config.groupId)
 			}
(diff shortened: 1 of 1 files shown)


samples/sampler/src/ThemeEnvironment/ThemeEnvironment.js
```diff
@@ -189,13 +190,20 @@ const StorybookDecorator = (story, config) => {
...
 			title={`${config.kind} ${config.story}`.trim()}
 			description={config.description}
 			locale={locale}
 			{...skinKnobs}
 			skinVariants={boolean('night mode', Config) && 'night'}
-			accent={color('accent', (!newSkin && accentFromURL ? accentFromURL : defaultColors[currentSkin].accent), Config.groupId)}
-			highlight={color('highlight', (!newSkin && highlightFromURL ? highlightFromURL : defaultColors[currentSkin].highlight), Config.groupId)}
+			accent={useSkinDefaultStyles && !allSkins ?
+				defaultColors[skinKnobs.skin].accent :
+				color('accent', (!newSkin && accentFromURL ? accentFromURL : defaultColors[currentSkin].accent), Config.groupId)
```
MikyungKim, line 201: Why didn't we use `skinKnobs.skin` in the first place instead of `currentSkin`?

ybsung: `currentSkin` is always `Gallium` now.

samples/sampler/src/ThemeEnvironment/ThemeEnvironment.js
```diff
@@ -176,6 +176,7 @@ const StorybookDecorator = (story, config) => {
 	const currentSkin = skinFromURL ? skinFromURL : Config.defaultProps.skin;
 	const newSkin = (memory.skin !== currentSkin);
 	memory.skin = currentSkin;  // Remember the skin for the next time we load.
+	const useSkinDefaultStyles = boolean('default skin styles', Config);
```
0x64, line 179: It looks better to re-order knobs so the new knob could be the next position of `skin` knob.

0x64: BTW, a name starting with `use` is okay? :)

0x64:
I'm wondering that `getPropFromURL` is really required for colors.

PR #296, opened 2020-07-20
PLAT-113850: Fixed broken logic covering edge cases of an invalid `data-index` value

0x64:
We're using an internal `getNumberValue` function to convert a string into a number for internal logic. The current implementation of `getNumberLogic` is much faster than `parseInt` but if a string is _undefined_ or _a string contains any character beside space_, it returns 0 instead of invalid value like NaN so we can't catch an exceptional case.
Also, in our usage, `getNumberValue` should return an invalid value like _-1_ or _NaN_ to test if a DOM node is an item or not since we use this function to check `data-index` attribute value of an item's DOM node.
To handle this issue, this PR updates `getNumberValue` with another fast implementation and makes it to return _-1_ for an invalid string.

### DIFF
--- a/VirtualList/useThemeVirtualList.js
+++ b/VirtualList/useThemeVirtualList.js
@@ -15,8 +15,12 @@ const SpotlightPlaceholder = Spottable('div');
 
 const
 	nop = () => {},
-	// using 'bitwise or' for string > number conversion based on performance: https://jsperf.com/convert-string-to-number-techniques/7
-	getNumberValue = (index) => index | 0;
+	getNumberValue = (index) => {
+		// using '+ operator' for string > number conversion based on performance: https://jsperf.com/convert-string-to-number-techniques/7
+		let number = +index;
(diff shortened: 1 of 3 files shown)


VirtualList/useThemeVirtualList.js
```diff
@@ -141,7 +147,7 @@ const useSpottable = (props, instances) => {
 		mutableRef.current.isScrolledBy5way = false;
 		mutableRef.current.isScrolledByJump = false;
 
-		if (nextIndex >= 0) {
+		if (nextIndex >= 0 && index >= 0) {
```
jeonghee27, line 150: I think we need this change on useEvent.js line 159 too.

jeonghee27: I got the reason why we cannot applying this on useEvent.js line 159.
the "if" statement structure is different with sandstone.

PR #299, opened 2020-07-22
FORTECH-89: Customizable button size for IncrementSlider

daniel-stoian-lgp:
Feature Added
Added new prop buttonSize( small/large). The increment/decrement buttons and slider padding will adjust based on this value.


IncrementSlider/IncrementSliderButton.js
```diff
@@ -46,12 +47,13 @@ const IncrementSliderButtonBase = kind({
...
 				onHold={onTap}
 				onHoldPulse={onTap}
+				size={size}
 			/>
 		);
 	}
 });
 
-const OnlyUpdate = onlyUpdateForKeys(['children', 'disabled', 'icon', 'spotlightDisabled', 'aria-label']);
+const OnlyUpdate = onlyUpdateForKeys(['children', 'disabled', 'icon', 'size', 'spotlightDisabled', 'aria-label']);
```
adrian-cocoara-lgp, line 56: Please sort alphabetically

IncrementSlider/IncrementSlider.js
```diff
@@ -103,6 +103,15 @@ const IncrementSliderBase = kind({
...
 		backgroundProgress: PropTypes.number,
 
+		/**
+		 * The size of the buttons.
+		 *
+		 * @type {('small'|'large')}
+		 * @default 'large'
+		 * @public
+		 */
+		buttonSize: PropTypes.oneOf(['small', 'large']),
```
adrian-cocoara-lgp, line 113: sort alphabetically

ybsung, line 113: We have the `decrementIcon` and the `incrementIcon` for the + and - buttons.
How about renaming `buttonSize` as `iconSize`?

samples/sampler/stories/default/IncrementSlider.js
```diff
@@ -17,6 +17,7 @@ storiesOf('Agate', module)
 		() => (
 			<IncrementSlider
 				active={boolean('active', Config)}
+				buttonSize={select('buttonSize', ['small', 'large'], Config, 'large')}
```
ybsung, line 20: We don't need to define the `large` default value because the story book already know it from Config.
`decrementIcon={select('decrementIcon', decrementIcons, Config)}`

ybsung: Could you omit `horizontal` in `orientation={select('orientation', ['horizontal', 'vertical'], Config, 'horizontal')}` below with the same reason above?

IncrementSlider/IncrementSlider.js
```diff
@@ -155,6 +155,15 @@ const IncrementSliderBase = kind({
...
 		focused: PropTypes.bool,
 
+		/**
+		 * The size of the icons.
+		 *
+		 * @type {('small'|'large')}
+		 * @default 'large'
+		 * @public
+		 */
+		iconSize: PropTypes.oneOf(['small', 'large']),
```
Djspaceg, line 165: Maybe this is a silly question, but why `iconSize` instead of just plain `size`? That way it would be more intuitive for a skin to define different values for the slider size (as done below anyway) in addition to the icon size. We'd have a more flexible component. Unless there are any objections, let's switch this over to just "size" and feed it directly to `Icon`. Don't forget to also create a `.large` class in the LESS file somewhere, so the packing process can obfuscate the "large" class too.

daniel-stoian-lgp: I will switch to "size".
Regarding the .large class in the less file, I don't see any component which has a defined less style for the default value of a prop. Should I just create one and leave it empty?

Djspaceg: You're right about the `.large` class not technically being necessary for the presentation of the component. However, this is a case where the theme (before your team was brought on) didn't include all of the necessary parts or was incorrectly configured. We've recently been looking into and tightening up the applied classNames and discovered that several components don't define all of the classes they use in their own CSS file. This causes a global class name to be applied to the component instead of a managed or obfuscated class name, which leads to a bug or class name leakage out of the component.

For example, a Button should have all of its local class names obfuscated, like the following:
* `Button_button__IquEL`
* `Button_button__19SR2`
* `neutral`
* `highContrast`
* `spottable`
* `Button_opaque__KzGtS`
* `Button_focusExpand__3Uo5q`
* `Button_iconBefore__17IJX`
* `Button_small__3ISz6`
* `Button_minWidth__3GYA2`
* `Button_minWidth__2W7ZL`
* `Button_selected__3lw9t`
* `Button_selected__2TuaD`
* `Button_small__zI8I_`
* `Button_small__3ISz6`

When we _don't_ include one of those class names in the CSS/LESS file, the list looks like this instead:
* `Button_button__IquEL`
* `Button_button__19SR2`
* `neutral`
* `highContrast`
* `spottable`
* `Button_opaque__KzGtS`
* `Button_focusExpand__3Uo5q`
* `Button_iconBefore__17IJX`
* `Button_small__3ISz6`
* `Button_minWidth__3GYA2`
* `Button_minWidth__2W7ZL`
* `Button_selected__3lw9t`
* `Button_selected__2TuaD`
* `large`

This `.large` class name is applied with no corresponding rule, which then leaves it as a global class, which causes the leakage.

What we can do in this case, instead of refactoring all of the `.large` styles into that class name, is create an empty rule-set with `/* this style of comment */` which won't get pruned out by the compiling process and allow the JS to read the class name, so it gets treated like the rest. Here's an example of where we've done it in Agate: https://github.com/enactjs/enact-theme-agate/blob/fe89db0a5d65f62dab661f82dc57751e466ed528/Button/Button.module.less#L74

This is a general practice that we need to be on the look-out for; it just hadn't come up before now. :) One of our team members here wrote a little script to detect these recently. I'll see if I can share it so it's easier to spot these in Agate.

daniel-stoian-lgp: Thank you for the detailed explanation. It now makes sense to us :) I pushed the empty rule-set

PR #300, opened 2020-07-22
PLAT-114689: Create a new icon font specifically for use with Silicon skin

Djspaceg:
A new font file (and supporting code) was added to support customized icons for Silicon skin.


ybsung:
image

In Agate Icon Sampler with `silicon` skin, the Scroll Up and Down Icon Buttons are broken.

Steps to reproduce
1. Launch Agate sampler
2. Select Icon
3. Knobs > Global Knobs > Skin > Silicon
4. Reduce the viewport height to show vertical Scrollbar

Djspaceg:
@teckliew , can we copy all of the glyphs that are in the agate-icons file and insert them at the same codepoints, then add them to the silicon icon list?

I think that will be the safest approach, so when someone switches to Silicon from some other skin, they're only getting _more_ icons rather than a slightly different set.

Djspaceg:
We were able to address the missing icons with a technique that actually layers the silicon icons onto the base Agate icons, so now even the icons follow the same logic and extensibility as the rest of the variables. Special thanks to @teckliew for finding and remapping the unique silicon icons to now unique codepoints to support this approach.

PR #268, opened 2020-06-05
FORTECH-38: Update styles for CheckboxItem

stanca-pop-lgp:
Feature Added
Updated styles for Agate CheckboxItem component.

Comments
Updated Agate Checkbox component styling according to zeplin design. Added unit tests.

Enact-DCO-1.0-Signed-off-by: Stanca Pop stanca.pop@lgepartner.com

PR #302, opened 2020-07-24
FORTECH-87: Item's bottom border has a slight angle at the extremities

adrian-cocoara-lgp:
Added fix for Item's bottom-border and RadioItem's checked color

### DIFF
--- a/styles/variables-silicon.less
+++ b/styles/variables-silicon.less
@@ -43,7 +43,7 @@
 // CheckboxItem
 // ---------------------------------------
 @agate-checkboxitem-icon-margin: 9px;
-@agate-checkboxitem-font-size: 30px;
+@agate-checkboxitem-font-size: 27px;
 
 // ColorPicker
 // ---------------------------------------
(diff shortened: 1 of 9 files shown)


styles/variables-silicon.less
```diff
@@ -130,14 +130,15 @@
...
-@agate-item-border-width:     3px;
-@agate-item-font-size:        27px;
-@agate-item-font-weight:      normal;
-@agate-item-height:           90px;
-@agate-item-height-large:     120px;
-@agate-item-label-font-size:  30px;
-@agate-item-side-outset:      30px;
-@agate-item-sub-text-size:    24px;
+@agate-item-border-width:        0;
+@agate-item-border-bottom-width: 3px;
```
ybsung, line 134: Because `box-sizing` is `border`, the Item width and height has reduced by removing the top, left, and right bottom width. Did you intent to do it? Could you check the GUI guide?

adrian-cocoara-lgp: I made a small change to `Item`'s height to correspond to GUI guide.

Djspaceg:
This is no longer necessary based on my changes (refactoring) of #268.

For future reference, `border` supports shorthand assignments when setting the rules individually. This was resolved in the other ticket by setting `@agate-item-border-width: 0 0 3px 0;` and not adding new variables or rules.

Djspaceg:
I didn't realize there was also apparently `RadioItem` color assignments in this ticket to fix the `Item` border.

PR #305, opened 2020-07-29
FORTECH-83: Dropdown expands under the expanded Dropdown

daniel-stoian-lgp:
**Feature added**
Dropdown closes  when a click is made outside the component


Dropdown/Dropdown.js
```diff
@@ -246,6 +248,61 @@ const DropdownBase = kind({
...
+		clickedOutsideDropdown = ({target}) => !this.node.contains(target)
+
+		// If a click happened outside the component area close the dropdown by forwarding the onClick from Toggleable.
+		handleClick = handle(
+			this.clickedOutsideDropdown,
+			forward('onClick')
+		).bindAs(this, 'handleClick')
+
+		render () {
+			const {...rest} = this.props;
```
ybsung, line 295: I think we don't need this line. You could pass `this.props` directly.
```
<Wrapped {...this.props} />
```

ybsung:
@daniel-stoian-lgp 
Great work! For now I don't have any other approach. Let's go ahead.

Djspaceg:
Did you consider using the `ui/FloatingLayer` (with invisible scrim) to make this work? If so, what did it not do that you needed?

`FloatingLayer` supports automatic z-index handling (most recent thing on top) and supports outside click capture. It's used in several components in Agate that could be used as examples.

daniel-stoian-lgp:
> Did you consider using the `ui/FloatingLayer` (with invisible scrim) to make this work? If so, what did it not do that you needed?
> 
> `FloatingLayer` supports automatic z-index handling (most recent thing on top) and supports outside click capture. It's used in several components in Agate that could be used as examples.

I tried to use FloatingLayer, but it occupies by default the entire screen and I would have had to set up some positioning styling for the dropdownList for each skin, which is pretty consuming.
If you can help with how to use FloatingLayer properly in this case, I would appreciate it. I could only find examples for ContextualPopup and Popup.


FloatingLayer

FloatingLayer2

ybsung:
> > Did you consider using the `ui/FloatingLayer` (with invisible scrim) to make this work? If so, what did it not do that you needed?
> > `FloatingLayer` supports automatic z-index handling (most recent thing on top) and supports outside click capture. It's used in several components in Agate that could be used as examples.
> 
> I tried to use FloatingLayer, but it occupies by default the entire screen and I would have had to set up some positioning styling for the dropdownList for each skin, which is pretty consuming.
> If you can help with how to use FloatingLayer properly in this case, I would appreciate it. I could only find examples for ContextualPopup and Popup.

Hello @daniel-stoian-lgp 

I don't have pretty much information about your question. But there might be an example to position the Dropdown's floating layer based on the Dropdown Button position.

The `TooltipDecorator` has a FloatingLayer, too. But its FloatingLayer opens based on the wrapped component. There is the `tooltipRelative` prop in it.
https://github.com/enactjs/agate/blob/develop/TooltipDecorator/TooltipDecorator.js#L177

If the `tooltipRelative` prop is true, then the `FloatingLayerBase` renders near the wrapped component.
```
				if (!tooltipRelative) {
					renderedTooltip = (
						<FloatingLayerBase open={this.state.showing} noAutoDismiss onDismiss={this.hideTooltip} scrimType="none" key="tooltipFloatingLayer">
							{renderedTooltip}
						</FloatingLayerBase>
					);
				}
```

Please let me know if it is helpful or not. If not, I'll check another example.

Djspaceg:
Regarding the floating layer concerns about having to update every skin, I don't think this needs to happen. There are some qualities of components that are universal. We try to build these into the UI layer, so the theme just has to worry about the visual customization and not the structure of the base components. FloatingLayer was not architected well with this in mind and requires more work on the theme's behalf to set up the constant parts. The scrim positioning can be hard-coded, I think, since it's a pretty universal feature.

So, yes, the floating layer fills the whole screen, but the contents aren't meant to. They should be assigned their sizes like normal and the floating layer is a way to vertically (z-index) stack them and provide a container for them to be positioned inside of.

If you're not really interested in pursuing this approach (which I think solves many of the issues you're seeing, in addition to the layering order), then we can just approve this PR and get it merged, otherwise, I'm happy to help review when the next version is ready. :)

daniel-stoian-lgp:
> Regarding the floating layer concerns about having to update every skin, I don't think this needs to happen. There are some qualities of components that are universal. We try to build these into the UI layer, so the theme just has to worry about the visual customization and not the structure of the base components. FloatingLayer was not architected well with this in mind and requires more work on the theme's behalf to set up the constant parts. The scrim positioning can be hard-coded, I think, since it's a pretty universal feature.
> 
> So, yes, the floating layer fills the whole screen, but the contents aren't meant to. They should be assigned their sizes like normal and the floating layer is a way to vertically (z-index) stack them and provide a container for them to be positioned inside of.
> 
> If you're not really interested in pursuing this approach (which I think solves many of the issues you're seeing, in addition to the layering order), then we can just approve this PR and get it merged, otherwise, I'm happy to help review when the next version is ready. :)

I think it's a good idea to do the FloatingLayer approach on a next task :) Thank you!

PR #310, opened 2020-08-04
PLAT-115302: Sync up VirtualList

MikyungKim:
This PR syncs up agate/VirtualList with sandstone/VirtualList.
Some of bug fixes from sandstone/VirtualList needs to be synced to agate/VirtualList.
Please refer to the ticket for details.

### DIFF
--- a/VirtualList/useThemeVirtualList.js
+++ b/VirtualList/useThemeVirtualList.js
@@ -1,7 +1,9 @@
 import Spotlight, {getDirection} from '@enact/spotlight';
 import Accelerator from '@enact/spotlight/Accelerator';
 import Pause from '@enact/spotlight/Pause';
+import {getTargetByDirectionFromElement} from '@enact/spotlight/src/target';
 import {Spottable} from '@enact/spotlight/Spottable';
+import utilDOM from '@enact/ui/useScroll/utilDOM';
 import React, {useCallback, useEffect, useRef} from 'react';
 
 import {dataIndexAttribute} from '../useScroll';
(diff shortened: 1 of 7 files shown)


VirtualList/useThemeVirtualList.js
```diff
@@ -350,32 +331,25 @@ const useThemeVirtualList = (props) => {
...
 	delete rest.focusableScrollbar;
-	delete rest.scrollContainerContainsDangerously;
 	delete rest.scrollAndFocusScrollbarButton;
+	delete rest.scrollContainerContainsDangerously;
+	delete rest.scrollContainerHandle;
 	delete rest.scrollContainerRef;
 	delete rest.scrollContentHandle;
 	delete rest.spotlightId;
-	delete rest.scrollContainerHandle;
+
```
jeonghee27, line 348: Why did you add an empty line?

jeonghee27:
I think We need a plan to better sync by separating the parts for Overscroll, scrollbar, and aria after this ticket.
I think we should discuss it.

VirtualList/useThemeVirtualList.js
MikyungKim: Oh, by mistake. I'll remove.

MikyungKim:
About `noScrollByDrag`, I think we removed the prop from agate because it's determined by `platform.touchscreen` inside of useScroll. I'm wondering why sandstone has this prop...

About DOM structure, I think we kept this because overscroll effect of agate has not been determined and unlike sandstone, the content is resized based on the scrollbar's existence so we didn't just touch anything from the original one.
We can discuss it later.

PR #318, opened 2020-08-14
FORTECH-102: Fixed focus on dropdown activator after selection

Djspaceg:
### The Issue
After making a selection from a dropdown menu, focus would be lost and unrecoverable.

### The Solution
Upon closing (hiding) the menu, the focus would remain inside the dropdown list menu popup, and since the list container was set to "self-only", spotlight wasn't allowed to leave that container. However, the container was only hidden, not destroyed, so the normal flow couldn't happen, causing Spotlight to just be stuck in an invisible box, like a real-life mime.

From @germboy:
> I added a `Spotlight.focus()`  bit in the dropdown transition component’s `onHide` method that will set focus back to the dropdown button/item when it closes.

I added several other aria props and ID from Sandstone to help with accessibility.

### DIFF
--- a/Dropdown/Dropdown.js
+++ b/Dropdown/Dropdown.js
@@ -18,6 +18,7 @@
 import {handle, forward, forProp} from '@enact/core/handle';
 import kind from '@enact/core/kind';
 import {extractAriaProps} from '@enact/core/util';
+import Spotlight from '@enact/spotlight';
 import SpotlightContainerDecorator from '@enact/spotlight/SpotlightContainerDecorator';
 import Changeable from '@enact/ui/Changeable';
 import Group from '@enact/ui/Group';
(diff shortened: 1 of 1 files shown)


ybsung:
> I added several other aria props and ID from Sandstone to help with accessibility.

Thanks for supporting the A11y features in Dropdown.
But we might not need the `ariaLabelledBy` prop.
If we have to support it, then the prop name could be `aria-labelledby`.
But there is no title in the Dropdown, so I think, we could remove it.

ybsung (approved):
LGTM

Dropdown/Dropdown.js
```diff
@@ -256,6 +272,14 @@ const DropdownBase = kind({
  * @public
  */
 const DropdownDecorator = compose(
+	SpotlightContainerDecorator({
+		enterTo: 'default-element',
+		preserveId: true
```
MikyungKim, line 277: Would you please explain why this config needed?

MikyungKim (approved):
LGTM

PR #320, opened 2020-08-21
PLAT-115997: Create Panels and Panel samplers

ybsung:
### Additional Consideration

- Move Scroller from `Theme` component to `PanelsBase`, if not, DOM elements are generated abnormally from Storybook app in samples/sampler/src/ThemeEnvironment/ThemeEnvironment.js
- Sort props in samples/sampler/.storybook/config.js like Sandstone
- Rename `Agate`, `AgateFullscreen` as `Theme`, `ThemeFullscreen` in samples/sampler/src/ThemeEnvironment/ThemeEnvironment.js like other themes
By doing it, it could easily maintain storybook code among theme libraries
- Rename `Agate` as `Environment` in samples/sampler/src/configure.js like other themes

### More description for samples/sampler/src/ThemeEnvironment/ThemeEnvironment.module.less patch

The <div style="position: relative; z-index: 0;"> DOM element is added by storybook in different place in the DOM tree depending on our `noScroller`, `noPanel`, and `noPanels`. But its height is 0. So we need to define several CSS selectors to set the height.

```
// additional selector on first div right after themeEnvironment to add min-height to display Panels content when Panels are used as a component inside storybook.
.themeEnvironment > div:first-child,
```
In `Panels` sampler with `noPanels` of true and `noScroller` of true,  this CSS selector is for `<div style="position: relative; z-index: 0;">`. The `:first-child` could be omitted. But it would be more safe.

image

```
// additional selector when noPanel option is used
.themeEnvironment main > div {
	width: 100%;
	height: 100%;
	position: relative !important;
}
```
In `Panel` sampler with `noPanel` of true and `noScroller` of true, this CSS selector is for `<div style="position: relative; z-index: 0;">`.

image


0x64 (approved):
LGTM

samples/sampler/stories/default/Panels.js
```diff
@@ -0,0 +1,75 @@
+/* eslint-disable react/jsx-no-bind */
+
```
MikyungKim, line 2: Let's remove this line

MikyungKim:
I'm seeing below warnings when I checked `all skins` in Panel sample. Please check if you could fix it.
```
Warning: React does not recognize the `spotlightId` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `spotlightid` instead. 

Warning:React does not recognize the `hideChildren` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `hidechildren` instead. 
```

MikyungKim (approved):
LGTM

PR #321, opened 2020-08-21
FORTECH-69: Update style for DateTimePicker

adrian-cocoara-lgp:
Updated styles for DateTimePicker

- added a new prop for picker to show a secondary set of values for 'increment' and 'decrement'


internal/DateComponentPicker/DateComponentPicker.js
```diff
@@ -55,10 +55,12 @@ const DateComponentPickerBase = kind({
 	render: ({children, max, value, ...rest}) => (
 		<Picker
 			{...rest}
+			css={css}
 			index={value}
 			max={max}
 			min={0}
 			orientation="vertical"
+			showSecondaryValues
```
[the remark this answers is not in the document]
adrian-cocoara-lgp: `showSecondaryValues` is actually a prop, but it had a default value of `true` and I forgot to update the sampler with the new prop. (that's why `Gallium` also had 2 sets of values)
In order to fix this, I modified the code so the 'secondary sets of values' only appear when `Silicon` skin is used. What do you think of this approach?

DatePicker/DatePicker.module.less
```diff
@@ -6,4 +6,13 @@
 	.year {
 		/* These should be available to the JS but have no associated rules at this time. */
 	}
+
+	.divider {
+		width: 3px;
+		height: 21px;
+		opacity: 0.4;
+		background-color: #ffffff;
```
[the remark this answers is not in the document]
ybsung: Or @agate-white.

DatePicker/DatePickerBase.js
```diff
@@ -212,45 +221,54 @@ const DatePickerBase = kind({
...
+										disabled={disabled}
+										key="year-picker"
+										max={maxYear}
+										min={minYear}
+										onChange={onChangeYear}
+										value={year}
+										width={4}
+									/>
+									{!isYearLast && <div className={css.divider} />}
+								</React.Fragment>
```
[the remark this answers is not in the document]
adrian-cocoara-lgp: Your approach works. I went ahead and applied your solution. Thanks!

DateTimePicker/DateTimePicker.module.less
```diff
@@ -5,11 +5,26 @@
 
 .dateTimePicker {
 	position: relative;
+
+	.pickerBackground {
```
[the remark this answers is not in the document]
adrian-cocoara-lgp: Yep 👍

DateTimePicker/DateTimePicker.module.less
```diff
@@ -5,11 +5,26 @@
...
+		position: absolute;
+		width: 95%;
+		z-index: -1;
```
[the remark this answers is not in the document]
adrian-cocoara-lgp: You're right. We don't need it.

internal/DateComponentPicker/DateComponentPicker.js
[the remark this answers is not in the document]
ybsung: > In order to fix this, I modified the code so the 'secondary sets of values' only appear when Silicon skin is used. What do you think of this approach?

Sounds good to me.

> showSecondaryValues is actually a prop, but it had a default value of true
My opinion is that we need to set `showSecondaryValues` to true at this line if the skin is equal to `silicon`.
```
showSecondaryValues={skin === 'silicon'}
```

Additional comment:
If possible, I think, it would be better to provide the pro defining the number of the values in a picker. It would be 3 by default and it would be 5 in `silicon` skin. If you have any concern, you could ignore this additional comment.

ybsung:
Hello @adrian-cocoara-lgp 
Could you add animation when changing a value? I mean if changing a value from 5 to 6, then 5 and 6 numbers transform to up.

adrian-cocoara-lgp:
> Hello @adrian-cocoara-lgp
> Could you add animation when changing a value? I mean if changing a value from 5 to 6, then 5 and 6 numbers transform to up.

Animation will be taken care of in this task: https://harmony.lge.com:8443/issue/browse/FORTECH-71

DatePicker/DatePicker.module.less
```diff
@@ -1,18 +1,48 @@
...
+//
+//	.divider {
+//		width: 3px;
+//		height: 21px;
+//		opacity: 0.4;
+//		background-color: #ffffff;
+//		display: inline-block;
+//		vertical-align: inherit;
+//	}
+//}
```
ybsung, line 48: This code might not used anymore. If right, could you remove it?

ybsung:
There is a warning in a console.
```
react.development.js:315 Warning: Each child in a list should have a unique "key" prop.

Check the render method of `TimePickerBase`. See https://fb.me/react-warning-keys for more information.
    in HourPicker (at TimePickerBase.js:242)
    in TimePickerBase (at DateTimeDecorator.js:205)
    in DateTimeDecorator (at Changeable.js:188)
    in Changeable (at I18nDecorator.js:182)
    in I18nContextDecorator (at Skinnable.js:154)
```

ybsung (approved):

ybsung:
Hi @adrian-cocoara-lgp 
I fixed minor issues according to my additional comments. If you have any concern, please let me know.

ybsung (approved):
LGTM

DatePicker/DatePicker.module.less
```diff
@@ -1,9 +1,29 @@
...
+			height: 21px;
+			opacity: 0.4;
+			background-color: white;
+			display: inline-block;
+			vertical-align: inherit;
+			position: absolute;
+			top: 50%;
+			transform: translateY(-50%);
+			left: calc(100% - 3px);
+		}
```
Djspaceg, line 23: These are really specific. Could this be generalized at all? It's got colors, measurements, everything.

Also, it's targeting everything except the last element? I assume to put lines between each of the elements. I usually like to avoid using `:not()` since it works so differently from the way people expect. This could be refactored to use `nth-child` instead or the `+` operator to target everything after the first element; positioning the line to the left, instead of the right of each of the elements.

It's also both positioned with absolute positioning and set to inline-block? But then it also has a `vertical-align`?

Djspaceg (changes requested):
If both date picker and time picker have similar styling, let's reposition the styling into a common component that they can both share.

DatePicker/DatePicker.module.less
daniel-stoian-lgp: removed  display and vertical-align. 
When it comes to repositioning the common styling, I moved ths style to DateTime.module.less. I hope it's the good way of doing :)

daniel-stoian-lgp:
@Djspaceg , ready for review

DatePicker/DatePicker.module.less
```diff
@@ -1,9 +1,11 @@
 // DatePicker.module.less
 //
+
 .datePicker {
 	.day,
 	.month,
 	.year {
 		/* These should be available to the JS but have no associated rules at this time. */
 	}
+
```
Djspaceg, line 10: Random newline not necessary. ;)

DatePicker/DatePickerBase.js
```diff
@@ -183,10 +185,12 @@ const DatePickerBase = kind({
 
 	styles: {
 		css,
+		dateTimeCss,
```
Djspaceg, line 188: I don't think this works the way you using it. You should be able to safely drop this line.

DatePicker/DatePickerBase.js
```diff
@@ -210,15 +214,15 @@ const DatePickerBase = kind({
 		delete rest.rtl;
 
 		return (
-			<DateTime {...rest}>
+			<DateTime {...rest} className={classnames(className, dateTimeCss.timePicker)}>
```
Djspaceg, line 217: This in not right.
` className={classnames(className, dateTimeCss.timePicker)}` can be dropped entirely, and `timePicker` can simply be added to the styles className block above, if you really want both `datePicker` and `timePicker` classes applied, which also seems really weird...

```javascript
	styles: {
		css,
		className: 'datePicker timePicker'
	},
```
It's more likely that you'll just want one class on the className:
```javascript
	styles: {
		css,
		className: 'timePicker'
	},
```

DatePicker/DatePickerBase.js
```diff
@@ -6,7 +7,8 @@ import $L from '../internal/$L';
 import {DateComponentRangePicker} from '../internal/DateComponentPicker';
 import DateTime from '../internal/DateTime';
 
-import css from './DatePicker.module.less';
+import css from '../internal/DateTime/DateTime.module.less';
+import dateTimeCss from '../internal/DateTime/DateTime.module.less';
```
Djspaceg, line 11: These import the same file. You can delete the second one (`dateTimeCss`).

DatePicker/DatePickerBase.js
```diff
@@ -1,3 +1,4 @@
+import classnames from 'classnames';
```
Djspaceg, line 1: This shouldn't be necessary; it's built into `kind()`.

DatePicker/DatePickerBase.js
```diff
@@ -210,15 +214,15 @@ const DatePickerBase = kind({
...
+			<DateTime {...rest} className={classnames(className, dateTimeCss.timePicker)}>
 				{order.map((picker) => {
 					switch (picker) {
 						case 'd':
 							return (
 								<DateComponentRangePicker
 									accessibilityHint={dayAccessibilityHint}
 									aria-label={dayAriaLabel}
-									className={css.day}
+									className={classnames(css.day, dateTimeCss.day)}
```
Djspaceg, line 225: This and the other two classnames changes below are redundant and unnecessary.

TimePicker/TimePickerBase.js
```diff
@@ -205,6 +207,7 @@ const TimePickerBase = kind({
 
 	styles: {
 		css,
+		dateTimeCss,
```
Djspaceg, line 210: This doesn't work. If you want this to work, it will need to be merged with the `css` object, but I'm reasonably certain this is not what you want to do here.

internal/DateTime/DateTime.module.less
```diff
@@ -16,3 +16,26 @@
...
+			content: '';
+			width: 3px;
+			height: 21px;
+			opacity: 0.4;
+			background-color: white;
+			position: absolute;
+			top: 50%;
+			transform: translateY(-50%);
+			right: calc(100% - 3px);
+		}
```
Djspaceg, line 39: This code block is far too specific to use. Please update this to the standard practice.

Djspaceg (changes requested):
There seems to be a large amount of messing with the classnames system. Please update these to use a more standard approach. I'm certain what you're attempting to do is possible within the existing mechanisms. Let's work together to get this refactored so it can do what you want and match the coding practices.

DatePicker/DatePickerBase.js
adrian-cocoara-lgp: Thanks for suggestion. 
I left it as `<DateTime {...rest} className={css.datePicker}>` and added `datePicker` in styles className since this module refers to `DatePicker`. I hope this is what you expected.

adrian-cocoara-lgp:
@Djspaceg Thanks for detailed explanation and suggestions. I pushed the changes, hopefully everything is ok now.

DatePicker/DatePickerBase.js
```diff
@@ -210,7 +210,7 @@ const DatePickerBase = kind({
 		delete rest.rtl;
 
 		return (
-			<DateTime {...rest}>
+			<DateTime {...rest} className={css.datePicker}>
```
Djspaceg, line 213: This isn't necessary. It's handled automatically by line 186.

TimePicker/TimePickerBase.js
```diff
@@ -236,7 +239,7 @@ const TimePickerBase = kind({
 		delete rest.rtl;
 
 		return (
-			<DateTime {...rest} css={css}>
+			<DateTime {...rest} css={css} className={classnames(className, dateTimeCss.timePicker)}>
```
Djspaceg, line 242: This shouldn't be necessary. The `css` override system should offer this functionality already, along with all of the below `classnames` usages, as long as the class names in the TimePicker's LESS file have the same name as the ones they're overriding.

internal/DateTime/DateTime.module.less
```diff
@@ -16,3 +17,31 @@
...
+
+		.applySkins({
+			&:nth-child(n+2)::before {
+				width: @agate-datetimepicker-divider-width;
+				height: @agate-datetimepicker-divider-height;
+				opacity: @agate-datetimepicker-divider-opacity;
+				background-color: @agate-datetimepicker-divider-bg-color;
+				right: calc(100% - @agate-datetimepicker-divider-width);
+			}
```
Djspaceg, line 44: What do you think about just merging these two blocks, for simplicity? Just do everything inside the `.applySkins` mixin.

Djspaceg (changes requested):
Still some work to do with the TimePicker, but a good start on the DatePicker.

Djspaceg:
Thanks for all the hard work! Glad we got a nice clean version of this component.

PR #322, opened 2020-08-21
PLAT-115231: Fix components to use base version for inline

ryanjduffy:
## Issue

Agate components using `inline()` are broken by enactjs/enact#2831

## Analysis

* `Icon` and `Heading` were importing the default export instead of the base export (because it didn't exist) and so were broken because the new default export is wrapped with `ForwardRef`.

## Resolution

Update to use base versions.


0x64 (approved):
:lgtm:

MikyungKim (approved):
LGTM

PR #324, opened 2020-08-24
Feature/PLAT-115997: Support accessibility features for Panels, TabbedPanels and TabGroup

ybsung:
Support A11y for Panels
: Fixed it based on Sandstone code

TabbedPanels
: Fixed next and previous tab button in the sampler

TabGoup
: Fixed the Tab including the ToggleButton wrapped with `div`


0x64 (approved):
LGTM

Header/Header.js
```diff
@@ -105,6 +113,11 @@ const HeaderBase = kind({
 	},
 
 	computed: {
+		'aria-label': ({title, subtitle}) => {
+			return subtitle ?
```
MikyungKim, line 117: It seems `aria-label` is public API but you replace the prop with this.
Did you intend this?

ybsung: I thought that we would support custom aria-label in another issue. The `aria-label` is not a  public API now. But the behavior has broken as you mentioned.

Anyway, as you mentioned it, I fixed it.

MikyungKim (approved):
:shipit:

PR #328, opened 2020-08-26
PLAT-115998: Add a11y support to Slider, SliderButton, and IncrementSlider

MikyungKim:
This PR adds a11y support to Slider, SliderButton, and IncrementSlider.
Slider and IncrementSlider added a function that focuses slider when the user starts to drag.
SliderButton added SliderButtonDecorator to read out changed values when a user drags.


seunghoh (changes requested):
Let's try with role="slider" and aria-valuetext rather than announce

seunghoh (approved):

seunghoh:
LGTM

SliderButton/SliderButtonBehaviorDecorator.js
```diff
@@ -0,0 +1,48 @@
+import platform from '@enact/core/platform';
+import PropTypes from 'prop-types';
+import React from 'react';
+import {findDOMNode} from 'react-dom';
+
+// Adds agate-specific SliderButton behaviors
+// * aria-valuetext handling
```
ybsung, line 7: The comment format is strange to me. Could you check it?

SliderButton/SliderButton.js
```diff
@@ -142,6 +144,7 @@ const SliderButtonBase = kind({
  */
 const SliderButtonDecorator = compose(
 	Pure,
+	SliderButtonBehaviorDecorator,
```
ybsung, line 147: Please update JSDoc for it.
```
* @mixes agate/SliderButtonBehaviorDecorator.SliderButtonBehaviorDecorator
```

SliderButton/SliderButtonBehaviorDecorator.js
```diff
@@ -0,0 +1,48 @@
...
+			// on platforms with a touchscreen, we want to focus slider when dragging begins
+			if (platform.touchscreen) {
+				findDOMNode(ref.current).focus(); // eslint-disable-line react/no-find-dom-node
+			}
+		}
+
+		return (
+			<Wrapped
+				aria-valuetext={valueText}
+				onChange={handleChange}
```
ybsung, line 29: Is it intended to add the `onChange` here? My expectation is before the `onDragStart`.

SliderButton/SliderButtonBehaviorDecorator.js
MikyungKim: I just copied from `SliderBehaviorDecorator`.

SliderButton/SliderButton.js
MikyungKim: 🆗

SliderButton/SliderButtonBehaviorDecorator.js
```diff
@@ -0,0 +1,48 @@
+import platform from '@enact/core/platform';
+import PropTypes from 'prop-types';
+import React from 'react';
+import {findDOMNode} from 'react-dom';
+
+// Adds agate-specific SliderButton behaviors
+// * aria-valuetext handling
+const SliderButtonBehaviorDecorator = (Wrapped) => {
+	const useValueText = ({onChange, ...rest}) => {
```
ybsung, line 9: Why does this component naming start with `use`? In my opinion, `useValueText ` is the component, not a hook.

```suggestion
const SliderButtonBehaviorDecorator = (Wrapped) => {
	// eslint-disable-next-line no-shadow
	function SliderButtonBehaviorDecorator ({onChange, ...rest}) {
```

By doing it, we don't have to define a displayName for it.

SliderButton/SliderButtonBehaviorDecorator.js
```diff
@@ -0,0 +1,48 @@
...
+		const {children} = rest;
+		const [valueText, setValueText] = React.useState(children[0]);
```
ybsung, line 11: Do we need to check if children is not undefined?
If getting the children texts from a server, it could be `undefined` at the first time.

MikyungKim: good point

PR #329, opened 2020-08-27
PLAT-115993: Updated a11y readout of SwitchItem

0x64:
SwitchItem reads checking status (checked/unchecked) along with a toggled status (on/off) but we just need a toggled status. To fix this, updated aria role from 'checkbox' to 'button'.


MikyungKim (approved):
LGTM

PR #330, opened 2020-08-27
PLAT-113933: Add a qa-a11y sample

ybsung:
### Add a qa-a11y sample to verify A11y behavior of the Agate components.

First of all, I cloned the qa-a11y sample in Sandstone.
And then I update files and Views properly for Agate APIs and UXs.
In addition, I fixed the following things
+ Use "react": "^16.8.0" and "react-dom": "^16.8.0" 
because Agate uses them.

+ All views except `ReadAlert` are now functional components, not React components 
to make their code simple

+ Define `useBoolArray` hook 
to manage open states of Popup and Drawer views.

In Sandstone: https://github.com/enyojs/sandstone/blob/develop/samples/qa-a11y/src/views/Popup.js, there are many state and handler definition.
```
	constructor (props) {
		super(props);
		this.state = {
			open0: false,
			open1: false,
			open2: false,
			open3: false,
			open4: false
		};
		this.handleOpen0 = this.handleOpen(0);
		this.handleOpen1 = this.handleOpen(1);
		this.handleOpen2 = this.handleOpen(2);
		this.handleOpen3 = this.handleOpen(3);
		this.handleOpen4 = this.handleOpen(4);
		this.handleClose0 = this.handleClose(0);
		this.handleClose1 = this.handleClose(1);
		this.handleClose2 = this.handleClose(2);
		this.handleClose3 = this.handleClose(3);
		this.handleClose4 = this.handleClose(4);
	}
	handleOpen = (expNum) => () => this.setState({['open' + expNum]: true});
	handleClose = (expNum) => () => this.setState({['open' + expNum]: false});

	const {open0, open1, open2, open3, open4} = this.state;
```

In Agate: https://github.com/enactjs/agate/blob/feature/PLAT-113933/samples/qa-a11y/src/views/Popup.js,
with `useBoolArray`, the code could be more concise.
```
	const [open, setOpen] = useBoolArray(5);
```
The open1, open2, open3, and open4 are replaced with `open` array and the `handleClose0` is replaced with setOpen(0, false).
 
+ Move all controls in the Panel's Body to in the Panel's Header
In Sandstone: 
image
In Agate
image


seunghoh (changes requested):
Leave some comments
- Let's have spaces between menu number and string
- Drawer: Can't close the drawer
- Dropdown: Agate Dropdown has a placeholder? 
- Header: Add some margin between button and horizontal line
- ImageItem: Image doesn't look 200 X 200
- Panels: Panel transition direction is opposite
- Picker: Check Vertical - Boston AIrport and Airport strings are overlapped
- Popup: Buttons has focus when popup opens (check with Buttons Popup)
- VritualGridList: Enable scrollbutton focused just like Scroller
- VirtualList: Enact scrollbutton focused

ybsung:
I fixed the following things based on the review.
- [x] means that I fixed the issue mentioned in the review.
- [ ] means that I didn't fixed it mentioned in the review because it's a component bug or not a but or something else.

> Leave some comments
> 
> * Let's have spaces between menu number and string
- [x]  : Fixed.

> * Drawer: Can't close the drawer

- [ ] Won't fix in the PR (A component issue). When opening a Drawer, the close button in the Drawer can't be spotted via 5-way keys. I hope that we could solve this issue in a separate issue. Until then, there is one way to spot the close button. After selecting an open button, please press 5-way up several times until the close button gets focus. If using a touch screen, then the close button can be clickable.

> * Dropdown: Agate Dropdown has a placeholder?
- [x] No. I fixed it.

> * Header: Add some margin between button and horizontal line
- [x] Fixed temporary. I don't think that it is the app issue, but a component issue. But I fixed it app side first. We need to revert it after fixing it component side. 

> * ImageItem: Image doesn't look 200 X 200
- [ ] Not a bug. It is the expected result because the ImageItem has a `horizontal` orientation prop. In this case, the ImageItem's width is 25% and the ImageItem's height is 100% of the item (not of the Image).
```
	&.horizontal {
		.image {
			width: 25%;
			height: 100%;
		}
	}
```  

> * Panels: Panel transition direction is opposite

- [ ] Not a bug. It's the expected result because if any arranger is not defined, then the default arranger is the `SlideRightArranger`. Panel transition direction is from left to right. So it's not a bug.

For more information, please check the https://github.com/enactjs/agate/blob/develop/Panels/Viewport.js#L80 code.

> * Picker: Check Vertical - Boston AIrport and Airport strings are overlapped

- [ ] Fixed temporary. Marquee has not been applied to a Picker. So the text in it could be overflow. For now, I just applied the `overflow: hidden`. But we need to discuss our policy for this case and to fix it properly.

> * Popup: Buttons has focus when popup opens (check with Buttons Popup)

- [ ]  Won't fix in the PR (A component issue). When showing a Popup, Spotlight doesn't move to any component in the Popup. So that's the reason why the buttons has focus. I think that we need to fix it component side in a separate issue. Until then, we could press a back key with a remote controller or a esc key in PC. If using a touch screen, then the close button can be clickable.

> * VritualGridList: Enable scrollbutton focused just like Scroller

- [x] Fixed.

> * VirtualList: Enact scrollbutton focused

- [x] Fixed.

ybsung:
I found more component bugs. I hope that we could fix them in separate issues.

- If not defining the width style in the `Native` CheckboxItem in the `VirtualList` view, the text in the CheckboxItem doesn't display. So I defined it app side. But this should be fixed component side.

Expected Result:
image

Actual Result
image

- When hovring the disabled Dropdown button, then the text in it doesn't display.
Actual Result
before hovering
image
after hovering
image

- Depending on the text width of the year Picker in DateTimePicker, the width of the year Picker changes
image

seunghoh:
@ybsung 
LGTM for this sample but please make components bug tickets that you mentioned above.
I created https://jira2.lgsvl.com/browse/PLAT-118480 for image item. Looks very wired with "200X200"

seunghoh (approved):

internal/Picker/Picker.module.less
```diff
@@ -32,6 +32,7 @@
 		.item,
 		.itemDecrement,
 		.itemIncrement {
+			overflow: hidden; // FIXME: Overflow text is clipped. But we need a better way.
```
0x64, line 35: Let's deal with this problem on another ticket.

0x64: > Picker: Check Vertical - Boston AIrport and Airport strings are overlapped

I checked @ybsung 's temporary fix resolves the text overlapping problem, but it looks better to expose the issue to be fixed properly in another ticket rather than a quick fix for component here without discussing about design strategies including a marquee.

PR #306, opened 2020-07-29
FORTECH-73: Keypad control

stanca-pop-lgp:
Feature Added
Created Agate Keypad component.

Comments
Added new Keypad component in Agate. Created styles for Gallium and Silicon skins, integrated component in StoryBook.

Enact-DCO-1.0-Signed-off-by: Stanca Pop stanca.pop@lgepartner.com

PR #334, opened 2020-08-31
PLAT-116058: Add a11y support to Keypad

MikyungKim:
Requires https://github.com/enactjs/agate/pull/306.
Please do not merge this PR before the above one merged.

This PR adds a11y support to Keypad.
It reads number for the number button, "call" for the phone button, and "backspace" for the back button.


Keypad/Keypad.js
```diff
@@ -161,18 +162,28 @@ const KeypadBase = kind({
 		return (
 			<Layout {...rest} align="center end" className={css.keypad} inline wrap>
 				{KEY_LIST.map((keyText, rowIndex) => {
+					const {icon, text} = keyText;
+					const isIcon = icon === 'arrowleftturn' || icon === 'phone';
+
+					let ariaLabel = text;
+					if (icon === 'arrowleftturn') {
+						ariaLabel = $L('Back Space');
```
jeonghee27, line 170: backspace ?

MikyungKim: 🆗

jeonghee27 (approved):
LGTM

Keypad/Keypad.js
```diff
@@ -111,6 +111,7 @@ const Key = kind({
 					css={css}
 					icon={children}
 					size="large"
+					role={null}
```
ybsung, line 114: the `role` would be before the `size`.

Keypad/Keypad.js
```diff
@@ -161,18 +162,28 @@ const KeypadBase = kind({
...
+							onKeyButtonClick={() => handleInputValue(isIcon ? icon : text)}
 							shrink
 							label={keyText.label}
-							text={keyText.icon === 'arrowleftturn' || keyText.icon === 'phone' ? null : keyText.text}
+							text={isIcon ? null : text}
 						>
-							{keyText.icon === 'arrowleftturn' || keyText.icon === 'phone' ? keyText.icon : null}
+							{isIcon ? icon : null}
 						</Cell>
 					);
```
ybsung, line 188: How about extracting some code in render() out of the kind? In my opinion, it would be enough to execute those code once. By doing it, we could simplify the `render` function.

```suggestion
const keyProps = KEY_LIST.map((keyText, rowIndex) => {
    const {icon, text} = keyText;
    const isIcon = icon === 'arrowleftturn' || icon === 'phone';

    let ariaLabel = text;
    if (icon === 'arrowleftturn') {
        ariaLabel = $L('backspace');
    } else if (icon === 'phone') {
        ariaLabel = $L('call');
    }

    return (
        aria-label={ariaLabel}
        component={Key}
        icon={isIcon ? icon : null}
        key={`key${rowIndex}-${text}`}
        shrink
        label={keyText.label}
        text={isIcon ? null : text}
    );
})

    render: ({disabled, handleInputValue, ...rest}) => {
        return (
            <Layout {...rest} align="center end" className={css.keypad} inline wrap>
                {keyProps.map(({icon, ...rest}) => {
                    return (
                        <Cell
                            {...props}
                            disabled={disabled}
                            onKeyButtonClick={() => handleInputValue(icon || text)}
                        >
                            {icon || null}
                        </Cell>
                    );
                }
            </Layout>
        );
    }
```

Keypad/Keypad.js
```diff
@@ -161,18 +162,28 @@ const KeypadBase = kind({
...
 				{KEY_LIST.map((keyText, rowIndex) => {
+					const {icon, text} = keyText;
+					const isIcon = icon === 'arrowleftturn' || icon === 'phone';
+
+					let ariaLabel = text;
+					if (icon === 'arrowleftturn') {
+						ariaLabel = $L('backspace');
+					} else if (icon === 'phone') {
+						ariaLabel = $L('call');
+					}
```
ybsung, line 173: I think that we could simplify the code above more.

```
const KEY_LIST = [
	...
	{icon: 'phone', label: 'call'},
	{icon: 'arrowleftturn', label: 'backspace'}
];
```

```suggestion
const {icon, label, text} = keyText;
const ariaLabel = icon ? $L(label) : text;
```

Keypad/Keypad.js
MikyungKim: I don't want to move this code from the render function. It lowers the readability.

Keypad/Keypad.js
MikyungKim: :okay:

ybsung (approved):
LGTM
Waiting for https://github.com/enactjs/agate/pull/306 being merged.

PR #342, opened 2020-09-11
PLAT-118160: Fix Checked message cut off in Dropdown

jeonghee27:
Enact-DCO-1.0-Signed-off-by: Jeonghee Ahn <jeonghee27.ahn@lge.com>

### Checklist

[Agate a11y]  Checked message cut off in Dropdown
When select a option of dropdown, the a11y message is different depending on the timing.

Expected Result: Provides a consistent A11y.. Read out `Option 2 checked Option 2 Button` or Read out `Option 2 Button` Always
Actual Result: Sometime Read out `Option 2 Button`. And Sometime `Option 2 checked Option 2 Button`

### Resolution
Sandstone is always read out  `Option 2 Button` . 
**So I fix the focus timing to onClose rather than onHide.**
This is also the same as Sandstone's implementation.

### Additional Considerations

### Links
PLAT-118160

### Comments


0x64 (approved):
:lgtm:

PR #343, opened 2020-09-11
PLAT-118165: Add a11y support to MediaPlayer

MikyungKim:
This PR adds a11y support to MediaPlayer.
If you move the knob on the slider, it will read "Jump to x minutes x seconds".

### DIFF
--- a/MediaPlayer/MediaPlayer.js
+++ b/MediaPlayer/MediaPlayer.js
@@ -4,6 +4,7 @@ import EnactPropTypes from '@enact/core/internal/prop-types';
 import kind from '@enact/core/kind';
 import {memoize} from '@enact/core/util';
 import {I18nContextDecorator} from '@enact/i18n/I18nDecorator';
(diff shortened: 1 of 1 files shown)


MediaPlayer/MediaPlayer.js
```diff
@@ -401,6 +413,14 @@ const MediaPlayerBehaviorDecorator = hoc((config, Wrapped) => { // eslint-disabl
 
 		onSliderChange = ({value}) => {
 			this.media.currentTime = value * this.state.duration;
+
+			const seconds = Math.floor(value * this.state.duration);
```
ybsung, line 417: Why didn't you use `this.media.currentTime` instead of `value * this.state.duration`?

MikyungKim: I don't want to read that variable since it's from media. I'd rather assign the value to a const variable.

0x64 (approved):

PR #345, opened 2020-09-14
FORTECH-105: Arc Component

daniel-stoian-lgp:
Added Arc component which lets the user generate a circular arc based on various props


samples/sampler/stories/default/Arc.js
```diff
@@ -0,0 +1,33 @@
...
+storiesOf('Agate', module)
+	.add(
+		'Arc',
+		() => (
+			<div style={{width:'40%', height:'40%'}}>
+				<Arc
+					color={select('color', prop.colors, Config)}
+					endAngle={number('endAngle', Config, {range: true, min: 0, max: 360})}
+					radius={number('radius', Config)}
+					startAngle={number('startAngle', Config, {range: true, min: 0, max: 360})}
```
adrian-cocoara-lgp, line 25: Maybe setting ` startAngle` and `endAngle` as simple inputs may be a better choice. It's easier for an user to just type the wanted value rather than selecting it form a range. What do you think?

adrian-cocoara-lgp:
Just a minor change, everything else looks good to me. Noice!

samples/sampler/stories/default/Arc.js
daniel-stoian-lgp: well, both of them have a lower limit and a higher limit( 0  - 360 ). That is why I chosed a range in the story

adrian-cocoara-lgp: I see. Ok, let's go with this approach.

adrian-cocoara-lgp (approved):
LGTM

Djspaceg:
This may be of interest, considering this component PR. 2 years ago I built a prototype set of components to build watch/clock faces, gauges, and dials in Enact. I built them into components and I worked out a bunch of interesting features for them: positioning the tick marks around a partial circle, orienting the numerals based on their position on the dial or relative to the inside/outside, etc. All this was built on top of ProgressBar, so the API would be simple as well. Below is the result of that. A base set of components, a theme set of components, and a sample app to tinker with the options.

1. https://github.com/enactjs/enact/compare/develop...feature/radial-components
2. https://github.com/enactjs/agate/compare/develop...feature/chronometer
3. https://github.com/enactjs/samples/compare/develop...feature/gauges

CombinedImages

_Try to ignore the bad color choices, these are all configurable with CSS._

Features include:
* Clockwise/counterclockwise
* Any size or quantity tick marks (on numbers)
* Custom labels for certain numerals
* Any sized dial, from 0 degrees to 360 degrees
   * At a full circle (360 degree range) the last value is automatically hidden, since it overlaps with the 0 position.
* Any starting angle
* Label orientations
   * "outward" - Top of the numeral is toward the center of the dial, with the bottom toward the outside edge
   * "inward" - Top of the numeral is closer to the edge, with the bottom toward the center
   * "horizon" - A combination of outward and inward, where the orientation flips below the center "horizon" line. Many clocks do this
   * "level" - Top of the numeral is always up, bottom is always down, like text in a book.

Screen Shot 2020-09-14 at 5 54 36 PM

ybsung:
Wow!! When did you develop such great components? Thanks for sharing your ideas that you did. I'll investigate your work more in depth to find out which ideas we could apply to Agate.

ybsung (changes requested):
Hello @daniel-stoian-lgp 
Thanks for your great work.

Overall it is well implemented as we expected. But I concerned about the value related with size.

1. The width and height of the `Arc`
https://github.com/enactjs/agate/pull/345/files#diff-d81650946d96b2fde1ca818f18c2d1f1R20
```
<div style={{width:'40%', height:'40%'}}>
```
It might make developers difficult to layout a `Arc` component with the DOM element above.
It would be better to set the `Arc`'s size based on the radius.
So if passing 100 radius prop, then the width and the height of the `Arc` will be 200px x 200px internally.

2. The `thickness` prop
If the `thickness` is 1 px, then stroke-width will be 1 px and the line width will be 0.5 px.
`thickness` naming could make me confuse because I thought the `thickness` is the line width incorrectly.
How about using `strokeWidth` instead the `thickness`?

3. If increasing the `thickness`, then the `Arc` size increases.
It changes the `Arc` layout. So I'd like to increase the `thickness` without changing the `Arc` size. It means the arc line increases to the center.

So I tried to make the patch for them. Could you check the commit https://github.com/enactjs/agate/commit/0ab45fd184a3ab4709152be3aa44b6da1c9c76e6?

In addition, I left some comments.

Arc/Arc.js
```diff
@@ -0,0 +1,142 @@
...
+		 * @type {String}
+		 * @default: #000000
+		 * @public
+		 */
+		color: PropTypes.string,
+
+		/**
+		 * The end angle(in degrees) of the arc.
+		 *
+		 * The value should be between 0 and 360 and should be greater than startAngle
```
ybsung, line 38: Add . at the end of the line.

Arc/Arc.js
```diff
@@ -0,0 +1,142 @@
+import kind from '@enact/core/kind';
```
ybsung, line 1: We need the module documentation like Button with @example at the first line.
```
/**
 * Sandstone styled button components and behaviors.
 *
 * @example
 * <Button>Hello Enact!</Button>
 *
 * @module sandstone/Button
 * @exports Button
 * @exports ButtonBase
 * @exports ButtonDecorator
 */
```

ybsung (changes requested):
@daniel-stoian-lgp 
Could you check my comments related with JSDocs?

Arc/Arc.js
daniel-stoian-lgp: I pushed  the fixes

ybsung (approved):

ybsung (approved):
LGTM

ybsung:
Hello @daniel-stoian-lgp 

For now, the current code is the exactly same one that we expected when defining this issue. But I didn't look into the Blake's commit - https://github.com/enactjs/agate/pull/345#issuecomment-692395952 yet. We might get any good insight from it. So it might take more time to merge this PR.

ybsung:
Here is the branch https://github.com/enactjs/agate/tree/feature/chronometer-yb including the following patches into Agate.

- enactjs/enact@develop...feature/radial-components
- develop...feature/chronometer
- enactjs/samples@develop...feature/gauges

To run the sample that Blake mentioned,
please run `npm i` and `npm run serve` in the `samples/gauges` folder.

ybsung:
I looked into the Blake's awesome output. There were lots of interesting features in it.

Related with this PR, `Ticks` component might give us brilliant ideas. Some arcs in the gauges were drawn with it.

There are several strength in the `Ticks`.
- Ticks are drawn with CSS styles
- Each Ticks could have their own event handlers.
- No complex math is not needed to draw an arc.

But there is one shortage in the `Ticks` for the PR. The tick shape is a quadrangle. So to draw an arc,
1. we could use several ticks for one arc, but the arc shape might be not good
2. we could draw a circle first and add ticks on the circle as the gap between arcs. But if using background image in an app, it might be a problem. So we need additional consideration for it.

Anyway, I think there is no reason not to merge this PR first while finding the way applying the `Ticks`'s good concepts to the `Arc`. Then we could refactor the `Arc` component later.

Arc/Arc.js
```diff
@@ -0,0 +1,153 @@
...
+ * @exports ArcDecorator
+ */
+
+import kind from '@enact/core/kind';
+import Pure from '@enact/ui/internal/Pure';
+import ri from '@enact/ui/resolution';
+import PropTypes from 'prop-types';
+import compose from 'ramda/src/compose';
+import React from 'react';
+import Skinnable from '../Skinnable';
```
MikyungKim, line 19: Would you please add one empty line above this?

daniel-stoian-lgp: sure

Arc/Arc.js
```diff
@@ -0,0 +1,153 @@
...
+ */
+const ArcBase = kind({
+	name: 'Arc',
+
+	propTypes: /** @lends agate/Arc.ArcBase.prototype */ {
+		/**
+		 * The color of the arc.
+		 *
+		 * @type {String}
+		 * @default: #000000
```
MikyungKim, line 42: I think we don't use `:` after `@default`. Please remove all of these.

ybsung:
Hello @daniel-stoian-lgp 

I want one more fix related with event handling.
While investigating the way to know which arc is clicked in the WindDirectionControl, I noticed that we need to pass a `onClick` prop to the `path` tag. And we need to handle `pointer-events` properly. Could you check my commit https://github.com/enactjs/agate/commit/86dbb1bf5beda8e8692f803afae34c0cacbfd102 based on the WindDirectionControl PR?

CC: @adrian-cocoara-lgp , @MikyungKim

Arc/Arc.js
```diff
@@ -0,0 +1,153 @@
...
+		 */
+		strokeWidth: PropTypes.number
+	},
+
+	defaultProps: {
+		color: '#000000',
+		endAngle: 310,
+		radius: 100,
+		startAngle: 50,
+		strokeWidth: 1
```
MikyungKim, line 93: Is this right? should it be 2?

daniel-stoian-lgp: it is only a default value. I can put 2 instead  :)

daniel-stoian-lgp:
> Hello @daniel-stoian-lgp
> 
> I want one more fix related with event handling.
> While investigating the way to know which arc is clicked in the WindDirectionControl, I noticed that we need to pass a `onClick` prop to the `path` tag. And we need to handle `pointer-events` properly. Could you check my commit [86dbb1b](https://github.com/enactjs/agate/commit/86dbb1bf5beda8e8692f803afae34c0cacbfd102) based on the WindDirectionControl PR?
> 
> CC: @adrian-cocoara-lgp , @MikyungKim

Thank you for the fix. We will test it right away and I will apply it to Arc.

Arc/Arc.js
MikyungKim: Oh, I think we could set this as `9` so it looks similar to the GUI guide.

ybsung:
@daniel-stoian-lgp 

@adrian-cocoara-lgp suggested that it would be better to provide the wider event handling area. To support it, we could add one additional path. So one for the visual part with pointer-events: none, the other for the event handle area part with pointer-events: auto, which stroke width is bigger then the `strokeWidth` prop. But please make sure that the latter path should not affect the Arc layout.

Arc/Arc.js
```diff
@@ -0,0 +1,173 @@
...
+		 * @default 50
+		 * @public
+		 */
+		startAngle: PropTypes.number,
+
+		/**
+		 * The stroke width of the arc.
+		 *
+		 * @type {number}
+		 * @default: 1
```
MikyungKim, line 92: Oh it should be `9`.

samples/sampler/stories/default/Arc.js
```diff
@@ -0,0 +1,31 @@
...
+};
+
+storiesOf('Agate', module)
+	.add(
+		'Arc',
+		() => (
+			<Arc
+				color={select('color', prop.colors, Config)}
+				endAngle={number('endAngle', Config, {range: true, min: 0, max: 360})}
+				radius={number('radius', Config)}
```
MikyungKim, line 23: How about add `onClick={action('onClick')}` here?

MikyungKim (approved):
LGTM

PR #347, opened 2020-09-16
PLAT-119500: Preserve opacity and color when hovering disabled components

ybsung:
### Issues

When hovering a disabled Dropdown Button, the text on it disappears

### Resolution

When hovering the focused and disabled components, we restored their natural opacity, then assign the faded text/content color. But it occurred more problems in Agate because there are several text colors. For example, some text colors are white and another text colors are black. So the faded text/content color could be white or black, then one of them can't be seen.

The Dropdown Button color is white and the hovered disabled text color is white because assigning the faded text/content color as I mentioned before.

So I hope we hold the disabled opacity and color.

### DIFF
--- a/ThemeDecorator/ThemeDecorator.module.less
+++ b/ThemeDecorator/ThemeDecorator.module.less
@@ -46,12 +46,6 @@
 	.disabled({
 		.vendor-opacity(@agate-disabled-opacity);
 
-		// Focused and disabled, must restore their natural opacity, then assign the faded text/content color
-		.focus({
-			.vendor-opacity(1);
-			color: @agate-spotlight-disabled-text-color;
-		});
-
 		cursor: default;
 
 		// Disabled children of disabled components
--- a/styles/colors-base.less
+++ b/styles/colors-base.less
@@ -72,7 +72,6 @@
 @agate-spotlight-color:                inherit;
 @agate-spotlight-bg-color:             inherit;
 @agate-spotlight-border-color:         inherit;
-@agate-spotlight-disabled-text-color:  inherit;
 @agate-spotlight-focus-bg-image:       inherit;
 
 // Button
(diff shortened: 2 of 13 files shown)


MikyungKim (dismissed):
:lgtm:

MikyungKim (changes requested):
I think we need to replace `1` to `@agate-disabled-opacity` rather than erase this.

MikyungKim (approved):
I've miss read something.. :lgtm:!

ThemeDecorator/ThemeDecorator.module.less
```diff
@@ -46,12 +46,6 @@
 	.disabled({
 		.vendor-opacity(@agate-disabled-opacity);
 
-		// Focused and disabled, must restore their natural opacity, then assign the faded text/content color
-		.focus({
-			.vendor-opacity(1);
-			color: @agate-spotlight-disabled-text-color;
```
jeonghee27, line 52: I am careful because this modification can affect all skins and component. actually, Silicon skin is affected and I cannot distinguish the disable status and disable with focus status.
I suggest some alternatives. Please let me know your opinion

1. Apply some style to Dropdown from Item.
There is no issue in the sampler. Because the Dropdown children is Item component.
Can't we just bring the Item component style into the DropDown style ?

2. Change the @agate-spotlight-disabled-text-color of Gallium.
current:   fade(@agate-spotlight-color, 40%); and @agate-spotlight-color is white.

3. Keep the opacity change.

		.focus({
			.vendor-opacity(1); // keep and change this value (option)
			// color: @agate-spotlight-disabled-text-color; // remove
		});

And, If you want revmoe " color: @agate-spotlight-disabled-text-color; "@agate-spotlight-disabled-text-color" is not used anymore.Please remove it

PR #348, opened 2020-09-16
FORTECH-107: DateTimePicker separator not correct for right-to-left locale

adrian-cocoara-lgp:
Added a fix for correctly positioning divider when using right-to-left locale


daniel-stoian-lgp (approved):
LGTM

internal/DateComponentPicker/DateComponentPicker.module.less
```diff
@@ -30,5 +29,19 @@
 			opacity: @agate-datetimepicker-divider-opacity;
 			background-color: @agate-datetimepicker-divider-bg-color;
 		}
+
+		&:nth-child(n+2)::before {
```
ybsung, line 33: `n + 2` with spaces would be better to me.

internal/DateComponentPicker/DateComponentPicker.module.less
```diff
@@ -30,5 +29,19 @@
...
+			content: "";
+		}
+
+		:global(.enact-locale-right-to-left) & {
+			&:nth-child(n+2)::before {
```
ybsung, line 38: I think that `n + 3`  would be better because
if not, the 2nd picker will be applied with the following 3 kinds of styles.
If using `n + 3`, then we will be applied with the first and the third styles in the following one.

```
.enact-locale-right-to-left .DateComponentPicker_dateComponentPicker__3e2Ns.gallium:nth-child(-n + 2)::before {
    content: "";
}
.enact-locale-right-to-left .DateComponentPicker_dateComponentPicker__3e2Ns.gallium:nth-child(n + 2)::before {
    content: none;
}
.DateComponentPicker_dateComponentPicker__3e2Ns.gallium:nth-child(n + 2)::before {
    content: "";
}
```

internal/DateComponentPicker/DateComponentPicker.module.less
```diff
@@ -30,5 +29,19 @@
...
+				content: none;
+			}
+
+			&:nth-child(-n+2)::before {
```
ybsung, line 42: Same above. Please use `-n + 2`.

adrian-cocoara-lgp:
Changes are pushed.

ybsung (approved):
LGTM

internal/DateComponentPicker/DateComponentPicker.module.less
```diff
@@ -30,5 +29,19 @@
 			opacity: @agate-datetimepicker-divider-opacity;
 			background-color: @agate-datetimepicker-divider-bg-color;
 		}
+
+		&:nth-child(n + 2)::before {
+			content: "";
+		}
+
+		:global(.enact-locale-right-to-left) & {
```
MikyungKim, line 37: You could use 
```suggestion
		.enact-locale-rtl({
```

daniel-stoian-lgp: thanks. suggested added

MikyungKim (approved):
:shipit:

PR #350, opened 2020-09-18
FORTECH-106: Added ArcSlider

daniel-stoian-lgp:
Added ArcSlider components


samples/sampler/stories/default/ArcSlider.js
```diff
@@ -0,0 +1,38 @@
...
+};
+
+storiesOf('Agate', module)
+	.add(
+		'ArcSlider',
+		() => (
+			<ArcSlider
+				backgroundColor={select('backgroundColor', prop.colors, Config)}
+				foregroundColor={select('foregroundColor', prop.colors, Config)}
+				endAngle={number('endAngle', Config, {range: true, min: 0, max: 360})}
```
stanca-pop-lgp, line 25: Please order alphabetically.

stanca-pop-lgp (approved):
LGTM

Arc/Arc.js
```diff
@@ -92,15 +92,37 @@ const ArcBase = kind({
...
+		 */
+		svgPointerEvents: PropTypes.oneOf(['none', 'auto']),
+
+		/**
+		 * A reference to the current svg.
+		 *
+		 * @type {Function}
+		 * @public
+		 */
+		svgRef: PropTypes.object
```
ybsung, line 112: For the `ref`, we usually name it as `componentRef`. Could you rename the `svgRef` as the `componentRef`?

Arc/Arc.js
```diff
@@ -109,12 +131,12 @@ const ArcBase = kind({
...
 	},
 
-	render: ({color, endAngle, onClick, radius, size, startAngle, strokeWidth, ...rest}) => {
+	render: ({children, color, endAngle, onClick, radius, size, startAngle, strokeWidth, svgPointerEvents, svgRef, ...rest}) => {
 		const halfStrokeWidth = strokeWidth / 2;
 		const viewBox = `-${halfStrokeWidth} -${halfStrokeWidth} ${radius * 2}  ${radius * 2}`;
 
 		return (
-			<svg {...rest} viewBox={viewBox} pointerEvents="none">
+			<svg {...rest} viewBox={viewBox} pointerEvents={svgPointerEvents} ref={svgRef}>
```
ybsung, line 139: How about `<svg pointerEvents="none" {...rest} viewBox={viewBox} ref={svgRef}>`?
Then we don't have to define `svgPointerEvents` prop, but app developers could override it.
CC: @MikyungKim

ArcSlider/ArcSlider.js
```diff
@@ -0,0 +1,249 @@
+/**
+ * Agate styled arc slider components and behaviors.
+ *
+ * @example
+ * <ArcSlider backgroundColor="blue" endAngle={200} foregroundColor="red" startAngle={0} radius={150} step={2} />
```
ybsung, line 5: The order would be 
```
radius={150} startAngle={0} step={2}
```

ybsung (changes requested):
Depending on the props, the Arc was drawn incorrectly.

Case1) When endAngle is smaller than startAngle
image

Case 2) When max = 0 and min = 10
image

Case 3) When changing the step from 10 to 100 with min = 0, max= 100, and value=90, the value is still 90. In this case, do we have to update the value to 0 or 100?
image

Arc/Arc.js
daniel-stoian-lgp: good idea

ArcSlider/ArcSlider.modules.less
```diff
@@ -0,0 +1,12 @@
+// ArcSlider.module.less
+//
+@import "../styles/mixins.less";
+@import "../styles/variables.less";
+@import "../styles/skin.less";
+
+.arcSlider {
+	.arc {
+		position: absolute;
+		overflow: visible;
```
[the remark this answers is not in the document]
daniel-stoian-lgp: I had to overwrite this style: 
svg:not(:root) {
    overflow: hidden;
}

ArcSlider/ArcSlider.js
```diff
@@ -0,0 +1,249 @@
...
+		min: PropTypes.number,
+
+		/**
+		 * Called when the mouse is down over the arc slider area.
+		 *
+		 * @type {Function}
+		 * @param {Object} event
+		 * @public
+		 */
+		onMouseDown: PropTypes.func,
```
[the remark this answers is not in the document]
daniel-stoian-lgp: I added support for touch now

daniel-stoian-lgp:
> Depending on the props, the Arc was drawn incorrectly.
> 
> Case1) When endAngle is smaller than startAngle
> image
> 
> Case 2) When max = 0 and min = 10
> image
> 
> Case 3) When changing the step from 10 to 100 with min = 0, max= 100, and value=90, the value is still 90. In this case, do we have to update the value to 0 or 100?
> image

Hello @ybsung, I need some guidance on what needs to be done for the cases you mentioned.
1. If endAngle < startAngle, then should an error be thrown? Or switch them in the logic(endAngle becomes startAngle). I assumed the user will put endAngle>startAngle
2. same for 1
3 This  happens also on the Slider component. 
The step is not taken into the calculations on initial rendering for Arc, only when calculating a new value onMouseDown/onTouchStart. 

Let me know your opinion

ybsung:
Hello @daniel-stoian-lgp 

I think that agate/ArcSlider and ui/Slider have similar functions but different looks, so it would be better to adapt any policy in ui/Slider to agate/ArcSlider.

> If endAngle < startAngle, then should an error be thrown? Or switch them in the logic(endAngle becomes startAngle). I assumed the user will put endAngle>startAngle
> same for 1

Let's add validate functions like ui/Slider. By doing it, we could show warnings in a console.
https://github.com/enactjs/enact/blob/develop/packages/ui/Slider/PositionDecorator.js#L143
```
			if (__DEV__) {
				const {min, value = min, max, step} = this.props;
				const props = {min, value: value || min, max, step};

				validateRange(props);
				validateStepValue(props);
				validateStepMax(props);
			}
```

> This happens also on the Slider component.
> The step is not taken into the calculations on initial rendering for Arc, only when calculating a new value onMouseDown/onTouchStart.

Let's leave as-is. Thanks for letting me know the behavior on the Slider.

Here are some more comments.
1. It would be better to use events in Touchable instead of using onMouseDown and onTouchStart to handle mouse events and touch events separately.
2. We need to forward any events if those events are passed to components.
3. It might be better to rename `calculateNewValue` as `emitChangeForPosition` because we need to forward `onChange` event like Slider, Picker, Input, Keypad having a value
Please check the following diff.
https://github.com/enactjs/agate/compare/feature/FORTECH-106...feature/FORTECH-106-yb

ArcSlider/ArcSliderBehaviorDecorator.js
```diff
@@ -0,0 +1,121 @@
...
+			super(props);
+
+			this.componentRef = React.createRef();
+
+			this.state = {
+				value: props.min
+			};
+		}
+
+		onMouseDown = (ev) => {
```
ybsung, line 34: We usually name it with `handle` for event handlers. So it might be `handleMouseDown`.
As I mentioned before, it's better to use `onDown` event in Touchable, then it would be `handleDown`.

ArcSlider/ArcSliderBehaviorDecorator.js
```diff
@@ -0,0 +1,121 @@
...
+			const componentRef = this.componentRef.current;
+
+			if (!platform.touchscreen) {
+				if (componentRef) {
+					componentRef.addEventListener('mousemove', this.calculateNewValue);
+					componentRef.addEventListener('mouseup', this.removeMouseListeners);
+				}
```
ybsung, line 41: If using `onDragStart`, `onDrag`, `onDragEnd` in `Touchable`, we don't need `addEventListener` and `removeEventListener`.

ArcSlider/ArcSliderBehaviorDecorator.js
```diff
@@ -0,0 +1,121 @@
...
+				this.calculateNewValue(ev);
+			}
+		};
+
+		onTouchStart = (ev) => {
```
ybsung, line 46: The `handleDown` as I commented above will cover this fuction.

ArcSlider/ArcSliderBehaviorDecorator.js
```diff
@@ -0,0 +1,121 @@
...
+			const componentRef = this.componentRef.current;
+
+			if (componentRef) {
+				componentRef.addEventListener('touchmove', this.calculateNewValue);
+				componentRef.addEventListener('touchend', this.removeMouseListeners);
+			}
```
ybsung, line 52: Same above.

ArcSlider/ArcSliderBehaviorDecorator.js
```diff
@@ -0,0 +1,121 @@
...
+				componentRef.removeEventListener('mousemove', this.calculateNewValue);
+				componentRef.removeEventListener('touchmove', this.calculateNewValue);
+				componentRef.removeEventListener('mouseup', this.removeMouseListeners);
+				componentRef.removeEventListener('touchend', this.removeMouseListeners);
+			}
+		};
+
+		// Calculates the new SVG value based on the mouse cursor coordinates and sets the new value into the state
+		calculateNewValue = (ev) => {
+			const {endAngle, max, min, radius, startAngle, step, strokeWidth} = this.props;
```
ybsung, line 69: It might be better to define this line after the line 74.

ArcSlider/ArcSliderBehaviorDecorator.js
```diff
@@ -0,0 +1,121 @@
...
+			if (step) {
+				const delta = (value - min) % step;
+				if (delta < step / 2) {
+					value -= delta;
+				} else {
+					value += step - delta;
+				}
+			}
+
+			this.setState({value: value});
```
ybsung, line 101: We need to forward `onChange` event when changing a value so that the App could know it.

samples/sampler/stories/default/ArcSlider.js
```diff
@@ -0,0 +1,39 @@
...
+		'ArcSlider',
+		() => (
+			<ArcSlider
+				backgroundColor={select('backgroundColor', prop.colors, Config)}
+				endAngle={number('endAngle', Config, {range: true, min: 0, max: 360})}
+				foregroundColor={select('foregroundColor', prop.colors, Config)}
+				max={number('max', Config)}
+				min={number('min', Config)}
+				onMouseDown={action('onMouseDown')}
+				onTouchStart={action('onTouchStart')}
```
ybsung, line 29: Instead of `onMouseDown` and `onTouchStart`, please add the following props.

```suggestion
				onChange={action('onChange')}
				onDown={action('onDown')}
				onDrag={action('onDrag')}
				onDragEnd={action('onDragEnd')}
				onDragStart={action('onDragStart')}
```

ybsung (approved):
LGTM

I'm happy to have such a great component in our Agate.

ArcSlider/ArcSlider.js
```diff
@@ -0,0 +1,244 @@
...
+		 * @public
+		 */
+		radius: PropTypes.oneOf([120, 150]),
+
+		/**
+		 * The start angle(in degrees) of the arc slider.
+		 *
+		 * The value should be between 0 and 360.
+		 *
+		 * @type {number}
```
MikyungKim, line 114:
```suggestion
		 * @type {Number}
```

ArcSlider/ArcSlider.js
```diff
@@ -0,0 +1,244 @@
...
+		 * @type {String}
+		 * @default 250
+		 * @public
+		 */
+		endAngle: PropTypes.number,
+
+		/**
+		 * The color of the foreground arc.
+		 *
+		 * @type {number}
```
MikyungKim, line 76:
```suggestion
		 * @type {Number}
```

ArcSlider/ArcSlider.js
```diff
@@ -0,0 +1,244 @@
...
+		 */
+		backgroundColor: PropTypes.string,
+
+		/**
+		 * Function that generates a reference to the arc svg.
+		 *
+		 * @type {Function}
+		 * @public
+		 */
+		componentRef: PropTypes.object,
```
MikyungKim, line 60: We have `EnactPropTypes.ref` for this kind.

ArcSlider/ArcSlider.js
```diff
@@ -0,0 +1,244 @@
...
+		 * @type {String}
+		 * @default #000000
+		 * @public
+		 */
+		backgroundColor: PropTypes.string,
+
+		/**
+		 * Function that generates a reference to the arc svg.
+		 *
+		 * @type {Function}
```
MikyungKim, line 57: should be `* @type {Object|Function}`

ArcSlider/ArcSlider.js
```diff
@@ -0,0 +1,244 @@
...
+		 * @type {number}
+		 * @default #0000ff
+		 * @public
+		 */
+		foregroundColor: PropTypes.string,
+
+		/**
+		 * The maximum value of the slider and should be greater than min.
+		 *
+		 * @type {number}
```
MikyungKim, line 85:
```suggestion
		 * @type {Number}
```

ArcSlider/ArcSlider.js
```diff
@@ -0,0 +1,244 @@
...
+		min: PropTypes.number,
+
+		/**
+		 * The radius of the arc circle.
+		 *
+		 * @type {Number}
+		 * @default 150
+		 * @public
+		 */
+		radius: PropTypes.oneOf([120, 150]),
```
MikyungKim, line 107: I think we could just get a number instead of this.

ArcSlider/ArcSlider.js
```diff
@@ -0,0 +1,244 @@
...
+		 * @type {Number}
+		 * @default 1
+		 * @public
+		 */
+		step: PropTypes.number,
+
+		/**
+		 * The stroke width of the arc slider.
+		 *
+		 * @type {number}
```
MikyungKim, line 132:
```suggestion
		 * @type {Number}
```

ArcSlider/ArcSlider.js
```diff
@@ -0,0 +1,244 @@
...
+		 * Defaults to the value of `min`.
+		 *
+		 * @type {Number}
+		 * @public
+		 */
+		value: PropTypes.number
+	},
+
+	defaultProps: {
+		backgroundColor: '#000000',
```
MikyungKim, line 150: For `backgroundColor` and `foregroundColor` default value,
I think these should be retrieved from "skin". WDYT?

daniel-stoian-lgp: how can we set a color based on skin on defaultProps?

ArcSlider/ArcSlider.js
```diff
@@ -0,0 +1,244 @@
...
+						r={ri.scaleToRem(15)}
+					/>
+				</Arc>
+			</div>
+		);
+	}
+});
+
+/**
+ * Applies Agate specific behaviors to [ArcSlider]{@link agate/ArcSlider.ArcSliderBase} components.
```
MikyungKim, line 209:
```suggestion
 * Applies Agate specific behaviors to [ArcSlider]{@link agate/ArcSlider.ArcSlider} components.
```

ArcSlider/ArcSlider.js
```diff
@@ -0,0 +1,244 @@
...
+ *
+ * @hoc
+ * @memberof agate/ArcSlider
+ * @mixes agate/Skinnable.Skinnable
```
MikyungKim, line 213:
```suggestion
  * @mixes ui/Touchable.Touchable
  * @mixes agate/Skinnable.Skinnable
```

ArcSlider/ArcSliderBehaviorDecorator.js
```diff
@@ -0,0 +1,126 @@
+import {forward} from '@enact/core/handle';
+import hoc from '@enact/core/hoc';
+import {validateRangeOnce, validateSteppedOnce} from '@enact/ui/internal/validators';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import {positionToAngle} from '../Arc/utils';
+import {angleToValue} from './utils';
```
MikyungKim, line 8: We need one empty line above.

Arc/Arc.js
```diff
@@ -45,6 +45,14 @@ const ArcBase = kind({
 		 */
 		color: PropTypes.string,
 
+		/**
+		 * A reference to the current svg.
+		 *
+		 * @type {Function}
+		 * @public
+		 */
+		componentRef: PropTypes.object,
```
MikyungKim, line 54: Oh this one also.

ArcSlider/ArcSlider.js
```diff
@@ -0,0 +1,246 @@
...
+		 * @public
+		 */
+		componentRef: EnactPropTypes.ref,
+
+		/**
+		 * The end angle(in degrees) of the arc slider.
+		 *
+		 * The value should be between 0 and 360 and should be greater than startAngle.
+		 *
+		 * @type {String}
```
MikyungKim, line 68:
```suggestion
		 * @type {Number}
```

ArcSlider/ArcSlider.js
MikyungKim: If you refer to `Picker`, you could get from `Skinnable({prop: 'skin'})` as "skin" prop and use this.
I guess we can't use it from defaultProps but from computed.

daniel-stoian-lgp: yes, it is possible to get it from computed, but that means the user will not be able to define these colors when using the Arcslider in another component

samples/sampler/stories/default/ArcSlider.js
```diff
@@ -0,0 +1,42 @@
...
+				onDragEnd={action('onDragEnd')}
+				onDragStart={action('onDragStart')}
+				radius={select('radius', prop.radius, Config)}
+				startAngle={number('startAngle', Config, {range: true, min: 0, max: 360})}
+				step={number('step', Config)}
+				strokeWidth={number('strokeWidth', Config)}
+			/>
+		),
+		{
+			text: 'The basic arc'
```
MikyungKim, line 40:
```suggestion
			text: 'The basic ArcSlider'
```

ArcSlider/ArcSlider.js
MikyungKim: 🤔 We could .. but let's just leave this as is.

Arc/Arc.js
```diff
@@ -45,12 +46,20 @@ const ArcBase = kind({
...
+		 */
+		componentRef: EnactPropTypes.ref,
+
 		/**
 		 * The end angle(in degrees) of the arc.
 		 *
 		 * The value should be between 0 and 360 and should be greater than startAngle.
 		 *
-		 * @type {number}
+		 * @type {Number}
```
MikyungKim, line 62: Thanks!

samples/sampler/stories/default/ArcSlider.js
```diff
@@ -0,0 +1,42 @@
...
+				onDragEnd={action('onDragEnd')}
+				onDragStart={action('onDragStart')}
+				radius={select('radius', prop.radius, Config)}
+				startAngle={number('startAngle', Config, {range: true, min: 0, max: 360})}
+				step={number('step', Config)}
+				strokeWidth={number('strokeWidth', Config)}
+			/>
+		),
+		{
+			text: 'The basic Arcslider'
```
MikyungKim, line 40: Sorry!
```suggestion
			text: 'The basic ArcSlider'
```

MikyungKim (approved):
👍 Nice job! 👏 :shipit:

PR #352, opened 2020-09-22
PLAT-120696: Add about page in sampler

seunghoh:
Add about page in the Agate Sampler


MikyungKim (approved):
💯

PR #354, opened 2020-09-25
PLAT-94190: Fix unresponsive VKB when touching Input component boundary

jeonghee27:
Enact-DCO-1.0-Signed-off-by: Jeonghee Ahn <jeonghee27.ahn@lge.com>

### Checklist

Agate - touching boundary area on Input component results in being unable to close VKB

### Resolution
* Root Cause
=> When touch not the <input> area on Input component, InputSpotlightDecorator delegate focus to <input>
but it was not preventDefault. so, engine focus again to clicked DOM.
During this series of processes, the VKB open and does not close, and does not work properly.
* Resolution
call ev.preventDefault to avoid focus the clicked DOM.

### Additional Considerations
The structure of InputSpotlightDecorator is too unnecessarily complex. And in the handlers of onBlur, onMouseDown, and onFocus move focus to other elements again.
When the focus delegate to another element, re-rendering occurs because of unnecessary state. it make analyzing very difficult .
I don't think this structure is desirable. So I think it's better to refactor this structure in the future.

### Links
PLAT-94190

### Comments


0x64 (approved):
The first commit fixes the issue and the second commit is a kind of improvement of life cycles. Commits look good to me, but it would be okay to revert the second commit if a merger does not agree to remove life cycle routines.

Looks good to me.

MikyungKim (approved):
LGTM

PR #355, opened 2020-09-25
FORTECH-61: Bug fixing for caption position overlay

adrian-cocoara-lgp:
Fixed position for caption when 'captionPosition'=overlay
Changed 'line-height' according to GUI design


daniel-stoian-lgp (approved):
Nice catch

ybsung (approved):
LGTM

MikyungKim:
@adrian-cocoara-lgp , Thanks for fixing! Please find my comments below.

The box that contains the overlay caption looks smaller than the image.
image

I'm not sure but shouldn't the caption starts from the bottom of the image? If the caption is short, it looks too above.
image

Should the overlayed caption marquee or not? WDYT?

adrian-cocoara-lgp:
Hello @MikyungKim 
Regarding your first comment about caption starting from bottom. I agree, for this I added a `min-height` and `max-height` for .captionOverlay
And the second, if we apply Marquee for overlayed caption, we won't be able to display caption on multiple rows.
What do you think?

MikyungKim (approved):
:lgtm:

PR #357, opened 2020-09-28
PLAT-119501: Fixed year picker width to not changing

MikyungKim:
This PR fixes changed year picker width when the value is changing.

### DIFF
--- a/internal/Picker/Picker.js
+++ b/internal/Picker/Picker.js
@@ -203,7 +203,24 @@ const PickerBase = kind({
 		 * @default 0
 		 * @public
 		 */
-		value: PropTypes.number
+		value: PropTypes.number,
+
+		/**
+		 * Choose a specific size for your picker. `'small'`, `'medium'`, `'large'`, or set to `null` to
+		 * assume auto-sizing. `'small'` is good for numeric pickers, `'medium'` for single or short
+		 * word pickers, `'large'` for maximum-sized pickers.
+		 *
+		 * You may also supply a number. This number will determine the minimum size of the Picker.
+		 * Setting a number to less than the number of characters in your longest value may produce
+		 * unexpected results.
+		 *
+		 * @type {('small'|'medium'|'large'|Number)}
+		 * @public
+		 */
+		width: PropTypes.oneOfType([
+			PropTypes.oneOf([null, 'small', 'medium', 'large']),
+			PropTypes.number
+		])
 	},
 
 	defaultProps: {
@@ -282,6 +299,7 @@ const PickerBase = kind({
 			step,
 			value,
 			valueId,
+			width,
 			...rest
 		} = props;
 		const currentValue = Array.isArray(values) ? values[value] : value;
@@ -296,8 +314,14 @@ const PickerBase = kind({
 		const decrementAriaLabel = `${currentValueText} ${decAriaLabel}`;
 		const incrementAriaLabel = `${currentValueText} ${incAriaLabel}`;
 
+		let sizingPlaceholder = null;
+		if (typeof width === 'number' && width > 0) {
+			sizingPlaceholder = <div aria-hidden className={css.sizingPlaceholder}>{'0'.repeat(width)}</div>;
+		}
+
 		delete rest.accessibilityHint;
 		delete rest['aria-valuetext'];
+		delete rest.orientation;
 
 		return (
 			<PickerRoot {...rest} onFlick={handleFlick}>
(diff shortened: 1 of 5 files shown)


internal/Picker/Picker.js
```diff
@@ -203,7 +203,24 @@ const PickerBase = kind({
...
+		 *
+		 * You may also supply a number. This number will determine the minimum size of the Picker.
+		 * Setting a number to less than the number of characters in your longest value may produce
+		 * unexpected results.
+		 *
+		 * @type {('small'|'medium'|'large'|Number)}
+		 * @public
+		 */
+		width: PropTypes.oneOfType([
+			PropTypes.oneOf([null, 'small', 'medium', 'large']),
```
ybsung, line 221: Are there any code to handle `'small', 'medium', 'large'` values in Agate Pickers like Sandstone? If not, let's support numbers. Then we might fix the qa-a11y Picker view using width prop of `large` and `medium`.

```
	determineClasses (css, decrementerDisabled, incrementerDisabled) {
		...

		return classnames(
			css.picker,
			css[orientation],
			css[width],
```

internal/Picker/Picker.js
```diff
@@ -296,8 +314,14 @@ const PickerBase = kind({
 		const decrementAriaLabel = `${currentValueText} ${decAriaLabel}`;
 		const incrementAriaLabel = `${currentValueText} ${incAriaLabel}`;
 
+		let sizingPlaceholder = null;
+		if (typeof width === 'number' && width > 0) {
+			sizingPlaceholder = <div aria-hidden className={css.sizingPlaceholder}>{'0'.repeat(width)}</div>;
+		}
```
ybsung, line 320: How about 
```
const sizingPlaceholder = (typeof width === 'number' && width > 0) ?
	<div aria-hidden className={css.sizingPlaceholder}>{'0'.repeat(width)}</div> :
	null;
```
?

internal/Picker/Picker.module.less
```diff
@@ -26,6 +25,14 @@
 		margin: @agate-picker-margin;
 		padding: @agate-picker-padding;
 
+		.sizingPlaceholder {
+			height: 0;
+			visibility: hidden;
+			padding-left: .extract(@agate-picker-padding, left)[];
+			padding-right: .extract(@agate-picker-padding, right)[];
```
ybsung, line 32: image
The year text has no padding.

image
But the `sizingPlaceholder` has the paddings. So the placeholder size will be bigger than the year text. Did you intend to do it?

internal/Picker/Picker.module.less
```diff
@@ -26,6 +25,14 @@
 		margin: @agate-picker-margin;
 		padding: @agate-picker-padding;
 
+		.sizingPlaceholder {
```
ybsung, line 28: image

`backgounrd-image` is applied to `.sizingPlaceholder `. Could you remove it?

ybsung (changes requested):
Could you find my comments?

In addition, it's out of scope, but if you can, could you delete `locale` and `value` props in the `DateTimeBase`?
```
const DateTimeBase = kind({
	name: 'DateTime',

	...
	render: ({children, css, ...rest}) => {
		delete rest.locale;
		delete rest.value;

		return (

			<div {...rest}>
				<div className={css.pickers}>
					{children}
				</div>
			</div>
		);
	}
});
```

internal/Picker/Picker.js
MikyungKim: We will support those values once animation related code got in soon.

internal/Picker/Picker.js
MikyungKim: Let's keep the original one.

internal/Picker/Picker.module.less
MikyungKim: 🆗

internal/Picker/Picker.module.less
MikyungKim: The width is the same as 90 + 12 +12 = 114.

ybsung (approved):
LGTM

jeonghee27:
image
the layout is broken. DId i do something wrong :(

MikyungKim:
> image
> the layout is broken. DId i do something wrong? :(

This happened because the width of DateTimePicker is a fixed value. We need to refactor Picker layout like sandstone's.
It will be handled in the picker animation ticket.

PR #361, opened 2020-10-05
FORTECH-115: Fix not to change Popup's title position when closeButton is visible

daniel-stoian-lgp:
fixed the change of position for Popup title when the closeButton is visible


adrian-cocoara-lgp (approved):
Noice! LGTM

ybsung:
image

If adding a long title text, the text is below the close button. Could you fix it together with proper paddings?

daniel-stoian-lgp:
> image
> 
> If adding a long title text, the text is below the close button. Could you fix it together with proper paddings?

Thanks for the notice. I added the fixes

ybsung:
> Thanks for the notice. I added the fixes

I verified it. Thanks.

ybsung (approved):
LGTM

styles/variables-base.less
```diff
@@ -447,6 +447,7 @@
 @agate-popup-border: 0 solid transparent;
 @agate-popup-padding: (@agate-component-spacing * 2);
 @agate-popup-closebutton-title-padding: 0 0 0 (@agate-button-small-height);
+@agate-popup-closebutton-centered-title-padding: @agate-popup-closebutton-title-padding;
```
jeonghee27, line 450: To fix for all skins, How about 
```
@agate-popup-closebutton-centered-title-padding: 0 (@agate-button-small-height);
```
and remove "@agate-popup-closebutton-centered-title-padding" in silicon.

Popup/Popup.module.less
```diff
@@ -84,6 +84,12 @@
 				padding: @agate-popup-closebutton-title-padding;
```
jeonghee27, line 84: I think we don't need this line.
this line make another bug in Gallium skin. 
the up and down padding is changed when Close button show.
Could you check this also? I saw this line was implemented by you.
There may be something I am missing.

styles/variables-silicon.less
```diff
@@ -238,7 +238,8 @@
 @agate-popup-border-radius: 39px;
 @agate-popup-border: none;
 @agate-popup-padding: (@agate-component-spacing * 2) 0 0 0;
-@agate-popup-closebutton-title-padding: 0 (@agate-button-small-height) 0 0;
+@agate-popup-closebutton-title-padding: 0 (@agate-button-small-height) 0 @agate-component-spacing;
```
jeonghee27, line 241: Please consider other skins. At least Gallium

Popup/Popup.module.less
daniel-stoian-lgp: yes we don;t need this line. Also, in order to avoid the top and bottom paadding change in Gallium, a 0 padding had to be set for the title

jeonghee27 (approved):
LGTM

PR #363, opened 2020-10-05
FORTECH-116: RadioItem - Hover style not the same as Zeplin

adrian-cocoara-lgp:
Hover over RadioItem bug fixed for Silicon skin to match Zepplin design
Changed default children text in sample from 'Sound' -> 'Hello RadioItem' so it would be more appropriate


daniel-stoian-lgp (approved):
LGTM

ybsung (approved):
LGTM

jeonghee27:
Could you modify the other skins with the same rule?

daniel-stoian-lgp:
> Could you modify the other skins with the same rule?

@jeonghee27, are we allowed to modify the other skins? Wouldn't it break some requirements for those skins?

jeonghee27:
> > Could you modify the other skins with the same rule?
> 
> @jeonghee27, are we allowed to modify the other skins? Wouldn't it break some requirements for those skins?

Of course your team can modify all other skins.
There is no GUI other than Gallium.
This obvious problem would be fine to apply to all.

PR #364, opened 2020-10-05
PLAT-123667: Visual transition for Picker components

stanca-pop-lgp:
Feature Added
Added visual transition for Picker components.

Comments
Added increased complexity to Picker components transitioning (iOS-like).


internal/Picker/Picker.js
```diff
@@ -290,39 +319,81 @@ const PickerBase = kind({
...
 			sizingPlaceholder = <div aria-hidden className={css.sizingPlaceholder}>{'0'.repeat(width)}</div>;
 		}
 
-		delete rest.accessibilityHint;
 		delete rest['aria-valuetext'];
+		delete rest.accessibilityHint;
 		delete rest.orientation;
 
+		const horizontal = orientation === 'horizontal';
+		let arranger = horizontal ? SlideLeftArranger : SlideTopArranger;
```
daniel-stoian-lgp, line 395: let's use const and move this line and the previous one above the "delete rest."  lines

daniel-stoian-lgp (approved):
LGTM

internal/Picker/PickerItem.js
```diff
@@ -0,0 +1,30 @@
...
+	name: 'PickerItem',
+
+	styles: {
+		css,
+		className: 'item',
+		publicClassNames: true
+	},
+
+	computed: {
+		className: ({children, styler}) => styler.append({numeric: !isNaN(Number(children))})
```
ybsung, line 18: `Number. isNaN` might be better instead of `isNan` to check if `Number(children)` is exactly `NaN` or not.

internal/Picker/PickerItem.js
```diff
@@ -0,0 +1,30 @@
+import kind from '@enact/core/kind';
+import React from 'react';
+
+import Marquee from '@enact/ui/Marquee';
+
+import css from './Picker.module.less';
+
+const PickerItemBase = kind({
```
ybsung, line 8: Please add JSDocs for `PickerItemBase` which might be `private` after fixing comments first because it is possible to remove the `PickerItemBase` component while refactoring the Picker in the PR.

internal/Picker/Picker.js
```diff
@@ -47,10 +50,19 @@ const secondaryDecrement = handleChange(-2);
  * @ui
  * @private
  */
-const PickerBase = kind({
+const PickerBase = kind( {
```
ybsung, line 53: Let's remove ` `.

internal/Picker/Picker.js
```diff
@@ -332,11 +403,21 @@ const PickerBase = kind({
 						aria-label={decrementAriaLabel}
 						className={css.secondaryItemDecrement}
 						disabled={isSecond}
-						onClick={handleSecondaryDecrement}
+						onClick={() => {
+							handleDecrement(); setTimeout(() => handleDecrement(), 150);
```
ybsung, line 407: Is there any reason that we have to use `setTimeout` here?

ybsung (commented):
Thanks for implementing the Picker virtual effect.

I have one concern. ui/ViewManger is not a lightweight component. If applying it to 5 PickerItems in one Picker, there are 5 ViewManagers in the Picker and interacting performance would be low. In addition, if dragging from one PickerItem to another PickerItem, it might be hard to track dragging events because the target DOM element changes and to move all PickerItems together. Is there any other better way like using only one ViewManager including 5 PickerItems?

@MikyungKim 
Please suggest any idea if you have.

internal/Picker/Picker.js
stanca-pop-lgp: Yes. If we don't set a timeout the 2 functions are called simultaneously and we can't see the double transition/animation.

internal/Picker/PickerItem.js
stanca-pop-lgp: We actually don't need this computed code now.

RangePicker/tests/RangePicker-specs.js
```diff
@@ -22,7 +22,7 @@ describe('RangePicker Specs', () => {
 
 	test('should increase by step amount on increment press', () => {
 		const picker = mount(
-			<RangePicker min={0} max={100} defaultValue={10} step={1} />
+			<RangePicker min={0} max={100} defaultValue={10} step={1} noAnimation />
```
ybsung, line 25: Could you order props alphabetically?

RangePicker/tests/RangePicker-specs.js
```diff
@@ -35,7 +35,7 @@ describe('RangePicker Specs', () => {
 
 	test('should decrease by step amount on decrement press', () => {
 		const picker = mount(
-			<RangePicker min={0} max={100} defaultValue={10} step={1} />
+			<RangePicker min={0} max={100} defaultValue={10} step={1} noAnimation />
```
ybsung, line 38: Same above.

internal/Picker/Picker.js
```diff
@@ -290,37 +319,79 @@ const PickerBase = kind({
...
 		const decrementAriaLabel = `${currentValueText} ${decAriaLabel}`;
 		const incrementAriaLabel = `${currentValueText} ${incAriaLabel}`;
 
+		const decrementValue = () => {
+			if (isFirst) {
+				return '';
+			} else if (Array.isArray(values)) {
+				return values;
+			} else {
+				return (<PickerItem key={clamp(min, max, value - step)} style={{direction: 'ltr'}}>{clamp(min, max, value - step)}</PickerItem>);
```
ybsung, line 351: Let's define a variable for `clamp(min, max, value - step)`.

internal/Picker/Picker.js
```diff
@@ -332,11 +403,21 @@ const PickerBase = kind({
...
+						}}
 					>
-						<div className={css.label}>
-							{isSecond ? '' : secondaryDecrementValue}
-						</div>
+						<ViewManager
+							aria-hidden
+							arranger={arranger}
+							className={css.viewManager}
+							duration={150}
```
ybsung, line 414: Let's define `150` as a constant variable so that we could tune is easily later.

internal/Picker/PickerItem.js
```diff
@@ -0,0 +1,34 @@
+import kind from '@enact/core/kind';
+import React from 'react';
+
+import Marquee from '@enact/ui/Marquee';
```
ybsung, line 4:
```suggestion
import kind from '@enact/core/kind';
import Marquee from '@enact/ui/Marquee';
import React from 'react';
```

internal/Picker/PickerItem.js
```diff
@@ -0,0 +1,34 @@
...
+
+import css from './Picker.module.less';
+
+/**
+ * Renders an Agate-styled Picker Item without any behavior.
+ *
+ * @class PickerItemBase
+ * @memberof agate/PickerItem
+ * @ui
+ * @public
```
ybsung, line 14: `PickerItemBase` is an internal component. So it might be better to define it as private.

internal/Picker/PickerItem.js
```diff
@@ -0,0 +1,34 @@
...
+
+import Marquee from '@enact/ui/Marquee';
+
+import css from './Picker.module.less';
+
+/**
+ * Renders an Agate-styled Picker Item without any behavior.
+ *
+ * @class PickerItemBase
+ * @memberof agate/PickerItem
```
ybsung, line 12: If the render function returns a component wrapped with any specific component, we always write it as @extends.
```
 * @extends ui/Marquee. Marquee
```

internal/Picker/PickerItem.js
```diff
@@ -0,0 +1,34 @@
...
+	},
+
+	render: (props) => (
+		<Marquee {...props} alignment="center" />
+	)
+});
+
+export default PickerItemBase;
+export {
+	PickerItemBase as PickerItem,
```
ybsung, line 32: There is no documentation for the PickerItem.

ybsung (approved):
LGTM

internal/Picker/Picker.js
```diff
@@ -406,6 +526,7 @@ const PickerBase = kind({
 const PickerDecorator = compose(
 	IdProvider({generateProp: null}),
 	Changeable,
+	Changeable({prop: 'reverseTransition'}),
```
0x64, line 529: I agree with this PR overall except data of `onChange` to an app. In this change, `onChange` brings `reverseTransition` value also to an app and it is not desired behavior. `reverseTransition` is used basically for `ViewManager` and not meaningful data to apps at all.

It looks better to filter the unexpected property between two `Changeable` HoCs.

internal/Picker/Picker.js
```diff
@@ -382,18 +493,47 @@ const PickerBase = kind({
...
+							{secondaryIncrementValue()}
+						</ViewManager>
 					</PickerButtonItem>
 				}
 			</PickerRoot>
 		);
 	}
 });
 
+const ChangeAdapter = hoc((config, Wrapped) => {
```
0x64, line 518: Could you please add comments for this new HoC? It looks better to add one for this even if it is private.

internal/Picker/Picker.js
```diff
@@ -405,7 +545,9 @@ const PickerBase = kind({
  */
 const PickerDecorator = compose(
 	IdProvider({generateProp: null}),
+	ChangeAdapter,
```
0x64, line 548: I expected that this new HoC would be placed between two `Changeable` HoCs since the upper `Changeable` has no need to get `reverseTransition` semantically. Could you please update this? It may be helpful for other developers to update this code later. :)

0x64 (commented):
The solution looks good to me. Just update comments and refine the code?

PR #367, opened 2020-10-06
FORTECH-117: Fixed  DatePicker/TimePicker broken layout for some locales

daniel-stoian-lgp:
Fixed  DatePicker/TimePicker broken layout for some locales(ta-IN)


stanca-pop-lgp (approved):
LGTM

DateTimePicker/DateTimePicker.module.less
```diff
@@ -34,7 +35,6 @@
 .applySkins({
 	.dateTimePicker {
 		font-size: @agate-picker-font-size;
-		width: @agate-datetimepicker-width;
```
ybsung, line 37: I think that the approach to fix the bug and most of the code looks good to me. But if reducing resolution, the DateTimePicker layout is broken. It might be related with this line. Do you have any idea to fix this issue?

image

daniel-stoian-lgp: I guess it is just too much content in the screen, for that particular resolution. The only thing that seems to fix it is to make the Picker padding 6px instead of 12px

ybsung (approved):
Now I couldn't reproduce the issue that I mentioned.
LGTM.

TimePicker/TimePickerBase.js
```diff
@@ -211,7 +211,7 @@ const TimePickerBase = kind({
 
 	computed: {
 		hasMeridiem: ({order}) => order.indexOf('a') >= 0,
-		meridiemPickerWidth: ({meridiem, meridiems}) => meridiems[meridiem].length * 2
+		meridiemPickerWidth: ({meridiem, meridiems}) => meridiems[meridiem].length
```
MikyungKim, line 214: Let's reduce this to 1.5.
Without the multiplier, the width varies too much when the meridem changes in am-ET locale.

MikyungKim (approved):
👍

PR #371, opened 2020-10-14
PLAT-122716: Fixed close button position for night mode

daniel-stoian-lgp:
Moved mixin inside applySkins() so it can be applied also for night mode


ybsung (approved):
LGTM

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+ 
+### Fixed
+- `agate/Panels` fixed children margin in controls for night mode
```
jeonghee27, line 8:
```suggestion
- `agate/Panels` to show close button properly for night mode
```
Because "fixed" is duplicated with ##Fixed, use do not type it.
And usually, changeLog is for App developer, so too detail things like technical way or implementation is not needed.

daniel-stoian-lgp: thanks for the suggestion and detailed explanation

PR #372, opened 2020-10-14
PLAT-122717: Fixed picker current value text position change in horizontal

daniel-stoian-lgp:

**Checklist**
 I have read and understand the contribution guide
 A CHANGELOG entry is included

**Issue Resolved / Feature Added**
Agate - Renamed classes .seCondaryItemIncrement and .secondaryItemDecrement to a common name
- Added the same width as .itemIncrement for the .secondaryItem  for horizontal picker to fix the descibed issue

**Links**
PLAT-122717


adrian-cocoara-lgp (approved):
LGTM

internal/Picker/Picker.module.less
```diff
@@ -95,7 +95,9 @@
 		&.horizontal {
 			.item,
 			.itemDecrement,
-			.itemIncrement {
+			.itemIncrement,
+			.secondaryItemIncrement,
+			.secondaryItemDecrement {
```
ybsung, line 100: Could we rename `secondaryItemIncrement` and `secondaryItemDecrement` as `secondaryItem` and use only `secondaryItem` for both secondary increment and decrement item? I think that `secondaryItemIncrement` and `secondaryItemDecrement` have the same rule.

ybsung:
We need to update [CHANGELOG.md](https://github.com/enactjs/agate/blob/develop/CHANGELOG.md) because of fixing our component.
```Suggestion
# Change Log

The following is a curated list of changes in the Enact sandstone module, newest changes on the top.

## [unreleased]
 
### Fixed
- `agate/Picker` to fix the picker item width in horizontal

## [1.0.0] - 2020-10-14

Initial release.
```

internal/Picker/Picker.module.less
daniel-stoian-lgp: good idea 👍

ybsung (approved):
LGTM

internal/Picker/Picker.module.less
MikyungKim: Oh, I'm sorry but I think it's not a good idea.
For future reference, I think the previous class names are clearer.
Would you please rollback the change you made here? I think the first commit looks good.

daniel-stoian-lgp: @MikyungKim ,I reverted this change.

PR #378, opened 2020-10-14
PLAT-122788: Fixed Icon list for incrementSlider sampler

daniel-stoian-lgp:
Updated the icon list in sampler app for increment/decrement buttons based on te current list of available icons in Agate


ybsung (approved):
LGTM

PR #381, opened 2020-10-16
PLAT-122893: Fixed SwitchItem icon position

daniel-stoian-lgp:

### Checklist

Agate - Fixed Switch icon position for the following cases:
a) RTL locale for Gallium/Silicon
b) both RTL and LTR locales for Titanium/Electro

### Resolution
* Root Cause
=> When locale is to ar-SA-Arabic, RTL and standard font, the switch icon is outside of the switch borders
 Solution -> add enact-locale-rtl() mixin
=> For Electro/Titanium skins, the above behavior happens also for LTR locale
 Solution -> update the switch icon css styling

### Links
PLAT-122893

### Comments


stanca-pop-lgp (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The following is a curated list of changes in the Enact agate module, newest cha
 - `agate/Panels` to show close button properly for night mode
 - `agate/LabeledIconButton` max-width so that huge sized icon will not be cut off
 - `agate/Spinner` to pause the animation when `paused` prop is true
+- `agate/Switch` icon position for all skins RTL locale and Electro/Titanium both LTR and RTL locale
```
ybsung, line 12: It might be better to replace `Switch` with `SwitchItem` because the `Switch` is a private component, app developers don't know it and they might use a `SwitchItem`.

ybsung (changes requested):
SwitchItem in RTL
image

We need to set margin right for RTL and margin left for LTR. For now there is margin left in RTL. So the Switch and the text sticks together.

SwitchItem/SwitchItem.module.less
```diff
@@ -9,7 +9,7 @@
 	}
 
 	.switchIcon {
-		margin-left: 10px;
+		.margin-start-end(9px, 0);
```
ybsung, line 12: Hello @daniel-stoian-lgp 
Is there any reason why there is 9px instead of 10px?

daniel-stoian-lgp: Yes, it's because of the resolution independence functionality. The rule is to use multiple of 3 values, for values above 3px

ybsung: Then it is reasonable to me.

ybsung (approved):
Thanks for fixing several issues.

LGTM

Switch/Switch.module.less
```diff
@@ -36,6 +40,11 @@
 			.icon {
 				-webkit-transition: left 0.2s, color 0.2s;
```
MikyungKim, line 41: We don't need `-webkit-` prefix since the build tool will do for us. Please remove these.

Switch/Switch.module.less
```diff
@@ -36,6 +40,11 @@
...
 				transition: left 0.2s, color 0.2s;
```
MikyungKim, line 42: Instead of below mixin, we could just use the current `transition` rule with `right` property added.
```
transition: left 0.2s, right 0.2s, color 0.2s;
```

Switch/Switch.module.less
```diff
@@ -49,6 +58,10 @@
 				:global(.spottable):focus & {
 					color: @agate-switch-selected-color;
 				}
+
+				.enact-locale-rtl({
+					right: @agate-switch-selected-icon-position-right;
```
MikyungKim, line 63: We don't need to add a variable for this. We could just use `@agate-switch-selected-icon-position-left`;
And you will need to add `position: absolute;` for `.icon` class since you are positioning from the right.

Actually, the result of the above code will revert the direction of the icon from
image
to
image.

According to https://developer.mozilla.org/en-US/docs/Archive/B2G_OS/Firefox_OS_apps/Firefox_OS_in_Arabic, the switch button should be mirrored when RTL so we'd like to change the direction.

Switch/Switch.module.less
```diff
@@ -27,6 +27,10 @@
 			font-size: @agate-switch-icon-font-size;
 			line-height: inherit;
 			margin: 0;
+
```
MikyungKim, line 30: You will need `position: absolute;` for this class.

SwitchItem/SwitchItem.module.less
MikyungKim, line 12: Let's make a variable for this `9px` like @agate-switch-icon-margin or so.

styles/variables-base.less
```diff
@@ -618,6 +618,7 @@
 @agate-switch-icon-font-size: (@agate-switch-height * @agate-switch-icon-scale);
 @agate-switch-icon-position-left: 0;
 @agate-switch-selected-icon-position-left: (@agate-switch-width - @agate-switch-height);
+@agate-switch-selected-icon-position-right: (@agate-switch-height - @agate-switch-width);
```
MikyungKim, line 621: Please remove this.

styles/variables-electro.less
```diff
@@ -164,6 +164,9 @@
 // Switch
 // ---------------------------------------
 @agate-switch-icon-scale: 2.8;
+@agate-switch-icon-position-left: -9px;
+@agate-switch-selected-icon-position-left: ((@agate-switch-width - @agate-switch-height) + @agate-switch-icon-position-left);
+@agate-switch-selected-icon-position-right: -(@agate-switch-height - @agate-switch-width + @agate-switch-icon-position-left);
```
MikyungKim, line 169: Please remove the right variable.

styles/variables-gallium.less
```diff
@@ -261,6 +261,7 @@
 @agate-switch-icon-scale: 3;
 @agate-switch-icon-position-left: -9px;
 @agate-switch-selected-icon-position-left: ((@agate-switch-width - @agate-switch-height) + @agate-switch-icon-position-left);
+@agate-switch-selected-icon-position-right: (@agate-switch-height - @agate-switch-width + @agate-switch-icon-position-left);
```
MikyungKim, line 264: Also, please remove this.

styles/variables-silicon.less
```diff
@@ -336,6 +336,7 @@
 @agate-switch-icon-scale: 3;
 @agate-switch-icon-position-left: -9px;
 @agate-switch-selected-icon-position-left: ((@agate-switch-width - @agate-switch-height) + @agate-switch-icon-position-left);
+@agate-switch-selected-icon-position-right: (@agate-switch-height - @agate-switch-width + @agate-switch-icon-position-left);
```
MikyungKim, line 339: same here.

styles/variables-titanium.less
```diff
@@ -65,6 +65,9 @@
 // Switch
 // ---------------------------------------
 @agate-switch-icon-scale: 2.8;
+@agate-switch-icon-position-left: -9px;
+@agate-switch-selected-icon-position-left: ((@agate-switch-width - @agate-switch-height) + @agate-switch-icon-position-left);
+@agate-switch-selected-icon-position-right: (@agate-switch-height - @agate-switch-width + @agate-switch-icon-position-left);
```
MikyungKim, line 70: same here.

MikyungKim (changes requested):
Good work, but we need to polish a little bit.
Let's make the switch icon's direction for RTL properly.
Most comments are from our @Djspaceg. Thank you!

MikyungKim:
And oh, don't forget to add screenshot test for this 👍

Switch/Switch.module.less
daniel-stoian-lgp: thanks for the Arabic guide 👍

MikyungKim (approved):
:shipit:

PR #396, opened 2020-11-03
PLAT-124516: Fixed console error when tapping SliderButton

adrian-cocoara-lgp:

### Checklist

* [x] A CHANGELOG entry](http://enactjs.com/docs/developer-guide/contributing/changelogs/) is included

Issue: A console error is shown when tapping SliderButton in QA-sampler

### Resolution
Solution: Refactored SliderButtonBehaviorDecorator and used `@enact/core/handle` to forward the `onChange` function 

### Links
PLAT-124516

### Comments


stanca-pop-lgp (approved):

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The following is a curated list of changes in the Enact agate module, newest cha
 - `agate/LabeledIconButton` max-width so that huge sized icon will not be cut off
 - `agate/Spinner` to pause the animation when `paused` prop is true
 - `agate/Picker` picker item width in horizontal for silicon
+- `agate/SliderButton` to not show console error in qa-sampler
```
ybsung, line 14: If possible, we recommend to order logs here alphabetically.

SliderButton/SliderButtonBehaviorDecorator.js
```diff
@@ -10,43 +12,71 @@ import {findDOMNode} from 'react-dom';
...
-		const ref = React.useRef();
-
-		function handleChange ({value}) {
-			onChange();
-			setValueText(children[value]);
+const SliderButtonBehaviorDecorator = hoc((config, Wrapped) => {
+	return class extends React.Component {
+		static displayName = 'SliderButtonBehaviorDecorator';
+
+		static propTypes = /** @lends agate/ArcSlider.ArcSliderBehaviorDecorator.prototype */ {
```
ybsung, line 19: It might not be updated after copy & paste.

```suggestion
		static propTypes = /** @lends agate/SliderButton. SliderButtonBehaviorDecorator.prototype */ {
```

SliderButton/SliderButtonBehaviorDecorator.js
```diff
@@ -10,43 +12,71 @@ import {findDOMNode} from 'react-dom';
...
-		onChange: PropTypes.func
+		render () {
+			return (
+				<Wrapped
+					{...this.props}
+					aria-valuetext={this.state.valueText}
+					onChange={this.handleChange}
+					onDragStart={this.handleDragStart}
+					ref={this.ref}
+					role="slider"
```
ybsung, line 74: We intended to add `aria-valuetext` and `role` before `{...rest}` because app developers could override them by passing props. It's the common way with A11y props. It breaks our coding rule, but it is easy to provide the way to override component default A11y features.

SliderButton/SliderButtonBehaviorDecorator.js
```diff
@@ -10,43 +12,71 @@ import {findDOMNode} from 'react-dom';
...
+
+			this.state = {
+				valueText: props.children ? props.children[0] : null
+			};
 		}
 
-		function handleDragStart () {
+		handleChange = ({value}) => {
+			this.setState(
+				() => ({valueText: this.props.children[value]}),
```
ybsung, line 49: Is there any reason to update the `valueText` state before calling `onChange` prop? I don't know why we didn't before exactly.

SliderButton/SliderButtonBehaviorDecorator.js
```diff
@@ -10,43 +12,71 @@ import {findDOMNode} from 'react-dom';
...
+				() => {
+					forward('onChange', {
+						type: 'onChange',
+						value
+					}, this.props);
+				}
+			);
+		};
```
ybsung, line 57: Is there any reason to call `forward` async? In other components, we just called it sync. So I'm curious about the needs for the async call.
```suggestion
		handleChange = ({value}) => {
			this.setState(() => ({valueText: this.props.children[value]}));
			forward('onChange', {
				type: 'onChange',
				value
			}, this.props);
		};
```

ybsung (changes requested):
Please find my comments.

SliderButton/SliderButtonBehaviorDecorator.js
adrian-cocoara-lgp: Uups. Thanks for noticing!

SliderButton/SliderButtonBehaviorDecorator.js
adrian-cocoara-lgp: Sorry, I don't quite understand your point here. Do you mean I should pass `{...this.props}` at the bottom like this?

```suggestion
					return (
				<Wrapped
					aria-valuetext={this.state.valueText}
					onChange={this.handleChange}
					onDragStart={this.handleDragStart}
					ref={this.ref}
					role="slider"
					{...this.props}
				/>
			);
```

SliderButton/SliderButtonBehaviorDecorator.js
adrian-cocoara-lgp: No particular reason, I saw it implemented the same way in other component. Do you think it should be changed?

SliderButton/SliderButtonBehaviorDecorator.js
adrian-cocoara-lgp: No reason as well. I changed it to call it sync. Thanks!

SliderButton/SliderButtonBehaviorDecorator.js
ybsung: My expectation is the following one.
Only a11y related props are defined before `{...this.props}`

```
		render () {
			return (
				<Wrapped
					aria-valuetext={this.state.valueText}
					role="slider"
					{...this.props}
					onChange={this.handleChange}
					onDragStart={this.handleDragStart}
					ref={this.ref}
				/>
			);
		}
```

SliderButton/SliderButtonBehaviorDecorator.js
ybsung: I'll discuss it with our members and let you know :)

SliderButton/SliderButtonBehaviorDecorator.js
adrian-cocoara-lgp: I see. Thank you!

SliderButton/SliderButtonBehaviorDecorator.js
ybsung: @MikyungKim , @0x64 , @jeonghee27 
Please share your opinions for the following points.

We usually call a props's even handler and setState() together in an event handler.
I want to discuss what is our preference for the following things.
- function vs object as the first argument in the SetState
My opinion:
If we need to refer the previous state or props or to call other functions after updating states, we need to use the function. In other cases, we could use the object simply.

- Which one between a prop's event handler and setState() should be called first?
My opinion:
If we call the `setState()` to update any status or any values which can be referred by apps, we need to call the prop's event handler after updating the state. Otherwise, I think we don't care including the case updating just a view.

### Original code
```
		function handleChange ({value}) {
			onChange();
			setValueText(children[value]);
		}
```
My opinion:
The `children` is cached as closer above and is not props or previous state. So it's ok to pass an object as the first argument of the `setValueText`.
I'm fine to call `onChange` first or after calling `setValueText`.

### The first patch
```
		handleChange = ({value}) => {
			this.setState(
				() => ({valueText: this.props.children[value]}),
				() => {
					forward('onChange', {
						type: 'onChange',
						value
					}, this.props);
				}
			);
		};
```
`(prevState, props) => ({valueText: props.children[value]}),` would be suitable than `() => ({valueText: this.props.children[value]}),` because of using proper props.
`forward` would be out of the `setState` because we called `setState` to update a view so that app developers could handle it early.

### The current patch
```
		handleChange = ({value}) => {
			this.setState(() => ({valueText: this.props.children[value]}));
			forward('onChange', {
				type: 'onChange',
				value
			}, this.props);
		};
```
If converting `this.setState(() => ({valueText: this.props.children[value]}));` to `this.setState((prevState, props) => ({valueText: props.children[value]}),);`, everything else is good to me.

0x64: > function vs object as the first argument in the SetState
I agree with @ybsung. If it is not required to access previous values, passing an object looks much better.

> Which one between a prop's event handler and setState() should be called first? 
In this case, `forward` handles _ev_ object anyway, so it looks not so important here. I don't have any preference of orders between them.

ybsung: Hi @adrian-cocoara-lgp 

Please check our discussion above and the following comment
> If converting this.setState(() => ({valueText: this.props.children[value]})); to this.setState((prevState, props) => ({valueText: props.children[value]}),);, everything else is good to me.

So my suggestion is..
```
		handleChange = ({value}) => {
			this.setState((prevState, props) => ({valueText: props.children[value]}));
			forward('onChange', {
				type: 'onChange',
				value
			}, this.props);
		};
```

adrian-cocoara-lgp: Thanks for detailed answers. I will change this accordingly.

ybsung (approved):
LGTM

MikyungKim (changes requested):
I don't know why we should change the component to the hoc.
I think we could just change `onChange` to
```
			forward('onChange', {
				type: 'onChange',
				value
			}, this.props);
```
will fix the error. You could replace this.props to props from by replacing `({onChange, ...rest})` to `props`.
Let's not converting this component to hoc.

adrian-cocoara-lgp:
> I don't know why we should change the component to the hoc.
> I think we could just change `onChange` to
> 
> ```
> 			forward('onChange', {
> 				type: 'onChange',
> 				value
> 			}, this.props);
> ```
> 
> will fix the error. You could replace this.props to props from by replacing `({onChange, ...rest})` to `props`.
> Let's not converting this component to hoc.

Considering this paragraph: "core/handle provides a set of utilities to support handling events for kind()s and React.Components" (https://enactjs.com/docs/modules/core/handle/), do you think we could use `forward` with a simple function component?

adrian-cocoara-lgp:
@MikyungKim I just tried your suggestion and it works. I will change it accordingly. Thanks!

MikyungKim:
> > I don't know why we should change the component to the hoc.
> > I think we could just change `onChange` to
> > ```
> > 			forward('onChange', {
> > 				type: 'onChange',
> > 				value
> > 			}, this.props);
> > ```
> > 
> > 
> > will fix the error. You could replace this.props to props from by replacing `({onChange, ...rest})` to `props`.
> > Let's not converting this component to hoc.
> 
> Considering this paragraph: "core/handle provides a set of utilities to support handling events for kind()s and React.Components" (https://enactjs.com/docs/modules/core/handle/), do you think we could use `forward` with a simple function component?

Actually, I didn't think of this much. I think we need a discussion on this. Thanks!

MikyungKim:
@adrian-cocoara-lgp I think we could use `forward` in functional components. Please go ahead changing.

MikyungKim (approved):
:shipit:

PR #406, opened 2020-11-16
PLAT-113577: Fix Input to not desctructure preventDefault

0x64:
## Issue

Invoking `preventDefault` from a destructured value caused a run time error

## Resolution

Remove `preventDefault` from destructuring and call on event object


ybsung (approved):
LGTM

PR #425, opened 2020-12-02
PLAT-125453: Replace TabbedPanels with Scroller for qa-a11y menu

ybsung:
Replace TabbedPanels with Scroller for qa-a11y menu because it is better to search and select a menu.


jeonghee27 (approved):
LGTM. but I prefer TabbedPanels.

PR #435, opened 2020-12-05
PLAT-123857: Support adding skins to the skin list in samplers via .env

ybsung:
We'd like to provide the way to add any additional skins via `.env` so that users could choose it among them in `Knobs > Global Knobs > skin`.


0x64 (approved):
:lgtm:

MikyungKim (approved):
:lgtm:

PR #436, opened 2020-12-07
PLAT-128519: FullscreenPopup supports `duration` prop as not only a string but also a number

ybsung:
FullscreenPopup supports `duration` prop as not only a string but also a number


jeonghee27 (approved):
LGTM

PR #439, opened 2020-12-08
PLAT-128492: Add UI tests and unit tests for SliderButton

adrian-cocoara-lgp:
### Checklist

* [ ] A CHANGELOG entry](http://enactjs.com/docs/developer-guide/contributing/changelogs/) is included

### Resolution
 - Added ui tests for SliderButton
 - Added unit tests for SliderButton

### Links
PLAT-128492

### Comments


daniel-stoian-lgp (approved):

SliderButton/tests/SliderButton-specs.js
```diff
@@ -0,0 +1,80 @@
+import {mount} from 'enzyme';
+import React from 'react';
+import SliderButton from '../SliderButton';
+
+const children = ['Light Speed', 'Ridiculous Speed', 'Ludicrous Speed'];
+
+describe('SliderButton Specs', () => {
+	test('should change value when clicking on the second button', () => {
```
jeongran1230, line 8: I think we can make it two different test cases.
One for emitting 'onChange' event and the other is related to set 'aria-valuetext'.
IMHO, having an assertion is good in a test method if possible. 
I don't need to prove onChange event is emitted but most of unit tests in enact have it. So, If you think it is necessary, please add it. 

```suggestion
	test('should emit an onChange event when changing value', () => {z
	const handleChange = jest.fn();
		const evt = {value: 1};
		const sliderButton = mount(
			<SliderButton onChange={handleChange}>
				{children}
			</SliderButton>
		);

		sliderButton.find('SliderButton').simulate('change', evt);

		const expected = 1;
		const actual = handleChange.mock.calls[0][0].value;
		expect(actual).toBe(expected);
    });

    test('should set "aria-valuetext" to value when value is changed', () => {
		const handleChange = jest.fn();
		const evt = {value: 1};
		const sliderButton = mount(
			<SliderButton onChange={handleChange}>
				{children}
			</SliderButton>
		);

		sliderButton.find('SliderButton').simulate('change', evt);

		const valueText = sliderButton.find('SliderButton').prop('aria-valuetext');
		expect(valueText).toBe('Ridiculous Speed');
	});
```

tests/ui/specs/SliderButton/SliderButtonPage.js
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	get self () {
+		return $(this.selector);
+	}
+	knobValue () {
+		return $(this.selector + ' .SliderButton_SliderButton_knob');
+	}
+	clickableItem (index) {
+		return $(this.selector + ` .SliderButton_SliderButton_track>div:nth-child(${index})`);
```
jeongran1230, line 17: How about the below code for finding clickable items? I think it is more clear but you have to change then index number overall.
```suggestion
	const parent = $(this.selector + ' .SliderButton_SliderButton_track');
        return = parent.$$('.SliderButton_SliderButton_client')[index];
```

tests/ui/specs/SliderButton/SliderButtonPage.js
```diff
@@ -0,0 +1,37 @@
...
+class SliderButtonInterface {
+	constructor (id) {
+		this.id = id;
+		this.selector = `#${this.id}`;
+	}
+
+	get self () {
+		return $(this.selector);
+	}
+	knobValue () {
```
jeongran1230, line 13: how about 'getKnob ' for method name?

jeongran1230:
@adrian-cocoara-lgp, Thank you for accepting my review. I a little bit changed the code.

PR #451, opened 2020-12-15
PLAT-129511: Added unit tests for Panels

daniel-stoian-lgp:

### Checklist

* [ ] A CHANGELOG entry](http://enactjs.com/docs/developer-guide/contributing/changelogs/) is included

### Resolution
 - Added ui tests for Panels

### Links
PLAT-129511
### Comments


stanca-pop-lgp (approved):

seunghoh (changes requested):
Can you add more test cases so that we can cover props as many as possible (ideally all props)? I understand you referred to Sandstone TCs and yes, we also need to update Sandstone TCs to cover all functions

samples/sampler/stories/default/Panels.js
```diff
@@ -54,6 +54,7 @@ const BasicPanels = () => {
 			noCloseButton={boolean('noCloseButton', Config, false)}
 			onApplicationClose={action('onClose')}
 			onBack={goPrevious}
+			orientation={select('orientation', ['horizontal', 'vertical'], Config)}
```
seunghoh, line 57: Good catch! 👍

seunghoh (approved):

PR #456, opened 2020-12-16
PLAT-129512: Add unit tests for Popup

stanca-pop-lgp:

### Checklist

### Resolution
- added extra unit tests for agate/Popup

### Links
PLAT-129512

### Comments


daniel-stoian-lgp (approved):

Popup/tests/Popup-specs.js
```diff
@@ -30,36 +44,82 @@ describe('Popup specs', () => {
...
-		const expected = false;
-		const actual = popup.prop('centered');
+			const expected = 'center';
+			const actual = popup.prop('position');
 
-		expect(actual).toBe(expected);
+			expect(actual).toBe(expected);
+		});
+
+		test('should not have center class', () => {
```
0x64, line 83: What about updating this to explain what is the key point here? If I understand correctly, this test case is to distinguish `position` prop's value 'center' and class 'center' by `centered` prop. Am I right? :)

stanca-pop-lgp: I did not explain it correctly, true. What I wanted to point out was that when it has position=center it has default styling, without adding an extra "center" class as it does when it has position=top.

stanca-pop-lgp: I think maybe we could just skip this test.

stanca-pop-lgp: I will delete this test for now and add it in this task: PLAT-129752: Sync Popup positioning with Sandstone.

PR #464, opened 2020-12-21
PLAT-129803: Fix LabeledIconButton shape to circular

MikyungKim:
This PR fixes LabledIconButton shape to circular.
And the warning saying 
```
Warning: React does not recognize the `iconOnly` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `icononly` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

### DIFF
--- a/LabeledIconButton/LabeledIconButton.js
+++ b/LabeledIconButton/LabeledIconButton.js
@@ -83,6 +83,14 @@ const LabeledIconButtonBase = kind({
 		 */
 		iconComponent: EnactPropTypes.component,
 
+		/**
+		 * True if button is an icon only button.
+		 *
+		 * @type {Boolean}
+		 * @private
+		 */
+		iconOnly: PropTypes.bool,
+
 		// forwarded from Spottable
 		pressed: PropTypes.bool,
 
@@ -122,6 +130,8 @@ const LabeledIconButtonBase = kind({
 		spriteCount,
 		...rest
 	}) => {
+		delete rest.iconOnly;
+
 		return UiLabeledIconBase.inline({
 			role: 'button',
 			...rest,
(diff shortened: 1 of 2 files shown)


LabeledIconButton/LabeledIconButton.js
```diff
@@ -130,6 +140,7 @@ const LabeledIconButtonBase = kind({
 					backgroundOpacity={backgroundOpacity}
 					icon={icon}
 					iconComponent={iconComponent}
+					iconOnly
```
jeonghee27, line 143: I thinks issue can be fixed by only this line.

LabeledIconButton/LabeledIconButton.js
```diff
@@ -83,6 +83,14 @@ const LabeledIconButtonBase = kind({
 		 */
 		iconComponent: EnactPropTypes.component,
 
+		/**
+		 * True if button is an icon only button.
+		 *
+		 * @type {Boolean}
+		 * @private
+		 */
+		iconOnly: PropTypes.bool,
```
jeonghee27, line 92: I don't know why we should add prop here. Do we have to add this for samplers?

LabeledIconButton/LabeledIconButton.js
MikyungKim: The issue could be fixed by this line but I'd like to fix the warning when the story loaded.

LabeledIconButton/LabeledIconButton.js
MikyungKim: As I was saying, to fix the warning, I need `delete rest.iconOnly;` line and it needs this propTypes for lint warning.

jeonghee27 (approved):
LGTM

PR #468, opened 2020-12-23
PLAT-129964: Replace internel utils.less with @enact/ui's less

0x64:
With the utils being migrated up to UI, they no longer need to be here.

### Resolution
This PR updates the LESS files to refer to the utils in their new home.


jeonghee27 (approved):
LGTM

PR #474, opened 2021-01-04
PLAT-122792: Fixed LabeledIconButton: Huge Icon is cut off

stanca-pop-lgp:

### Checklist

Issue: agate/LabeledIconButton: Huge size icon with label position `after` or `right` is cut off. 

### Resolution
Solution: Removed marquee from LabeledIconButton, adjusted component max-with.

### Links
PLAT-122792

### Comments


LabeledIconButton/LabeledIconButton.js
```diff
@@ -12,12 +12,15 @@
 
 import EnactPropTypes from '@enact/core/internal/prop-types';
 import kind from '@enact/core/kind';
+import Spottable from '@enact/spotlight/Spottable';
+import {ButtonDecorator as UiButtonDecorator} from '@enact/ui/Button';
 import {LabeledIconBase as UiLabeledIconBase, LabeledIconDecorator as UiLabeledIconDecorator} from '@enact/ui/LabeledIcon';
+import Pure from '@enact/ui/internal/Pure';
```
daniel-stoian-lgp, line 18: sort imports

daniel-stoian-lgp (approved):

ybsung:
There is one regression.
If passing null string (''), then the icon sticks to the left.

AS-IS
image

TO-BE
image

LabeledIconButton/LabeledIconButton.js
```diff
@@ -163,7 +166,10 @@ const LabeledIconButtonBase = kind({
  */
 const LabeledIconButtonDecorator = compose(
 	UiLabeledIconDecorator,
-	ButtonDecorator
+	Pure,
+	UiButtonDecorator,
+	Spottable,
+	Skinnable
```
ybsung, line 172: `ButtonDecorator` is replaced with 
```
	Pure,
	UiButtonDecorator,
	Spottable,
	Skinnable
```

By doing it, I'm worried that it is ok to remove `IconButtonDecorator`, `TooltipDecorator({tooltipDestinationProp: 'decoration'})`.

CC: @jeonghee27

jeonghee27: If LabeledIconButton don't need the iconOnly prop, IconButtonDecorator is not needed.
But if the tooltip support has not changed, let's keep TooltipDecorator.

stanca-pop-lgp:
@ybsung Icon sticks to the left when there are no children because LabeledIconButton has max-width: min-content;. Is there a reason why a LabeledIconButton should take more space than its icon if there are no children?
More than that, is there a reason why a LabeledIconButton would have no label? 
Thank you.

ybsung:
Hello @stanca-pop-lgp 

> @ybsung Icon sticks to the left when there are no children because LabeledIconButton has max-width: min-content;. Is there a reason why a LabeledIconButton should take more space than its icon if there are no children?
More than that, is there a reason why a LabeledIconButton would have no label?
Thank you.

The reason why I mentioned before is

First of all, if there is no special reason, it might be better to preserve component behavior and look. There were some space to the left before. But I found that there is no space in develop branch, too. So it would be better to fix it in a separate issue.

Second, as you mentioned, the LabeledIconButton  is usually used with a label, if without a label, the IconButton might be used. But app developers could pass the `children` prop asynchronously. So in my opinion, it might be better to keep the Button layout even though some props has been arrived yet or aren't passed.

In addition, it would be helpful to check any JS doc parse errors. I pushed one commit to fix some errors including one which is not related with your PR.

jeonghee27 (commented):
LGTM

PR #475, opened 2021-01-04
PLAT-129753: Sync Heading with Sandstone

stanca-pop-lgp:

### Checklist

Issue: agate/Heading: Sync Heading with Sandstone

### Resolution
Solution: Added marquee to Heading when text is too long.

### Links
PLAT-129753

### Comments


daniel-stoian-lgp (approved):
LGTM

Heading/Heading.js
```diff
@@ -108,11 +111,19 @@ const HeadingBase = kind({
...
  * @mixes agate/Skinnable.Skinnable
  * @public
  */
 const HeadingDecorator = compose(
+	setPropTypes({
+		marqueeOn: PropTypes.oneOf(['hover', 'render'])
+	}),
+	defaultProps({
+		marqueeOn: 'render'
+	}),
```
ybsung, line 124: My understanding for these lines is..

the `marqueeOn` prop in `packages/ui/Marquee/MarqueeDecorator.js` has `focus` as a default value and could have `hover`, `render`, and `focus`. But the `Header` can't be focused. So the default should be another value except `focus` and the possible values would be `render` and `hover`.

This PR provides the way to select one value between `hover` and `render` nicely instead of choosing one component side. But we need to make a decision what the UX is correct aligning with next Agate UX.

In addition, there is a new way to define defaultProp and propTypes from recompose witth `defaultProps` and `setPropTypes` in our components. I think this approach is quite useful to define HoC.

I like these ways. So LGTM. But I want to hear merger's opinion.

P.S: No doc parse error

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ The following is a curated list of changes in the Enact agate module, newest cha
 - `agate/ArcSlider` prop `aria-valuetext` to override `aria-valuetext` for it
 - `agate/Button` prop `iconOnly`, `iconPosition` and `minWidth`
 - `agate/DateTimePicker` prop `dayAriaLabel`, `hourAriaLabel`, `meridiemAriaLabel`, `minuteAriaLabel`, `monthAriaLabel`, and `yearAriaLabel`
+- `agate/Heading` `marquee` to text
```
jeonghee27, line 13:
```suggestion
- `agate/Heading` prop `marqueOn`  to select marquee condition
```

@MikyungKim @0x64 
How do you think?

Heading/Heading.js
jeonghee27: I agree that having the choice of "hover" and "render" is appropriate for `heading`.
And I believe this is suitable for the next Agate UX as well.
I've confirmed that the way to define props exactly same as Sandstone. I like to maintain this identity. So LGTM too.
@0x64  @MikyungKim Do you have any other opinions?

jeonghee27:
I Changed the changelog.

CHANGELOG.md
jeonghee27: ### Added

- `agate/Heading` prop `marqueOn` to determine when marquee begins

### Changed

- `agate/Heading` to support marquee

PR #480, opened 2021-01-05
PLAT-129901: Fixed Option text marquees unexpectedly

daniel-stoian-lgp:
### Checklist

Issues: 
agate/Button:  The icon is centered when the button has `minWidth`. This breaks the UX guide https://zpl.io/bPkLp3k 
agate/Dropdown: The dropdown  has agate/Marquee applied twice for silicon skin. Once directly and once indirectly via the inner agate/Button

### Resolution
Solution: 
- agate/Button: centered the text inside Button and  because of the UX specs: https://zpl.io/bPkLp3k
- added `flex-grow: 1` to avoid the icon to be centered. see UX specs
- agate/Dropdown silicon skin: Used agate/Button`s  iconPosition functionality; Removed all the custom Marquee related decoration because they come now via agate/Button
- `@agate-dropdown-button-icon-margin-right` was added in order to respect the ux guide regarding the position of the icon inside the dropdown button
- `@agate-dropdown-button-icon-width` - was added to avoid the position change of the icon when the marquee animation is ongoing;

### Links
PLAT-129901

### Comments


stanca-pop-lgp (approved):

adrian-cocoara-lgp (approved):

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ The following is a curated list of changes in the Enact agate module, newest cha
 ### Fixed
 
 - `agate/Dropdown` to show focused icon color for Silicon skin
+- `agate/Dropdown` to now show double marquee text
```
jeonghee27, line 16: now? not?Please do careful internal review. 
@stanca-pop-lgp @adrian-cocoara-lgp

Dropdown/Dropdown.js
```diff
@@ -46,6 +46,7 @@ import componentCss from './Dropdown.module.less';
 
 const oppositeDirection = {left: 'right', right: 'left', up: 'down', down: 'up'};
 const ContainerDiv = SpotlightContainerDecorator({enterTo: 'last-focused'}, 'div');
+const Button = Skinnable(ButtonBase);
```
jeonghee27, line 49: You also removed other decorator . 

Pure,
	IconButtonDecorator,
	TooltipDecorator({tooltipDestinationProp: 'decoration'}),
	UiButtonDecorator,
	Spottable

Is it intended? @adrian-cocoara-lgp @stanca-pop-lgp @daniel-stoian-lgp 
Because of this, I can't reach the dropdown with the 5way key in silicon.

I think we need to fix around 291 Line. But before that, DrowDown doesn't sync up with sandstone?

CHANGELOG.md
daniel-stoian-lgp: sorry for that

Dropdown/Dropdown.js
daniel-stoian-lgp: You are right. I added the decorators below. IconButtonDecorator was not added since there is use case for a button with icon for dropdown.

Pure,
UiButtonDecorator,
TooltipDecorator({tooltipDestinationProp: 'decoration'}),
Spottable,
Skinnable

daniel-stoian-lgp: we will analyze the sync of Dropdown very soon! 👍

Dropdown/Dropdown.js
```diff
@@ -35,17 +38,26 @@ import compose from 'ramda/src/compose';
...
 const ContainerDiv = SpotlightContainerDecorator({enterTo: 'last-focused'}, 'div');
+const ButtonDecorator = compose(
+	Pure,
+	UiButtonDecorator,
+	TooltipDecorator({tooltipDestinationProp: 'decoration'}),
+	Spottable,
+	Skinnable
+);
+const Button = ButtonDecorator(ButtonBase);
 const MarqueeButton = MarqueeDecorator({className: componentCss.marquee}, Button);
```
jeonghee27, line 61: I think there were two choices.

Can you explain why you are remove the marquee from the Button, not line 49? (Please leave on the description)

Because we are in remote, so please explain in detail.

daniel-stoian-lgp: On line 49, a class `marquee` is applied. it's role is to prevent the overlap of the marquee text with the arrow down/up icon on its' right. That's why I chosed to remove the marquee which comes from agate/Button.

If I were to choose the other one, some adjustments are needed on agate/Button. (add there a marquee class when applying MarqueeDecorator, then set flex-grow:1 for marquee and text-align:center)

Which approach would you prefer?

jeonghee27: I don't have any preference. if there is a reason, and an internal review decided that this direction is good, it's okay.

stanca-pop-lgp: The second approach sounds cleaner. I would opt for that.

adrian-cocoara-lgp: Actual approach looks good to me.

stanca-pop-lgp (approved):
LGTM

CHANGELOG.md
```diff
@@ -14,8 +14,10 @@ The following is a curated list of changes in the Enact agate module, newest cha
 
 ### Fixed
 
+- `agate/Button` to not center the icon when it has `minWidth`
```
jeonghee27, line 17: Something went wrong during the merge process.

Button/Button.module.less
```diff
@@ -318,6 +318,18 @@
 			}
 		}
 
+		// override Marquee's alignment.
+		.marquee {
+			text-align: center;
```
jeonghee27, line 323: After DrowDown sync, let's revisit if this part is not needed.

jeonghee27 (approved):
LGTM

seunghoh:
LGTM. Thanks for the detailed PR description. It helps to review a lot 👍

PR #487, opened 2021-01-13
PLAT-129911: Dropdown: Fixed initial focus when open dropdown with 5-way key

stanca-pop-lgp:

### Checklist

* [x] A CHANGELOG entry](http://enactjs.com/docs/developer-guide/contributing/changelogs/) is included

Issue: Initial focus is needed when opening dropdown with 5-way key.

### Resolution
Solution: Added `handleTransitionShow` function to add initial focus when opening dropdown with 5-way key. Fixed UI tests accordingly.

### Links
PLAT-129911

### Comments


Dropdown/Dropdown.js
```diff
@@ -299,7 +319,7 @@ const DropdownBase = kind({
 			{},
 			{},
 			{},
-			{childComponent: Item, itemProps: {size: 'small'}},
+			{childComponent: Item, itemProps: {size: 'small', css}, selectedProp: 'selected'},
```
daniel-stoian-lgp, line 322: I think that if you don't apply `css` to the child Item, you will not need to add all the .item styling in Dropdown.module.less and this will not affect the fix for the bug described in this PR

Dropdown/Dropdown.module.less
```diff
@@ -54,6 +54,21 @@
...
+				border-width: @agate-item-border-width;
+
+				.focus({
+					border-color: @agate-item-focus-border-color;
+				});
+
+				&[aria-checked="true"] {
+					&:not(:focus) {
+						color: @agate-dropdown-selected-color;
+					}
```
daniel-stoian-lgp, line 69: Also, this breaks the UX guide https://zpl.io/29nBwyA  select item keeps the same color

stanca-pop-lgp: Yes, this is true. Thank you for noticing.

Dropdown/Dropdown.js
stanca-pop-lgp: I had to apply it, as Item has accent colored border for selected. And I had to overwrite some style.

adrian-cocoara-lgp (approved):

stanca-pop-lgp:
This PR is obsolete.

PR #506, opened 2021-02-02
PLAT-123306: [Agate] TemperatureControl: Arc does not look a normal shape

paul-beldean-lgp:
### Checklist

- ArcSlider would not display correctly when 'min' was greater than 'max' or when 'min' and 'max' were equal

### Resolution
- Made ArcSlider display correctly
- The change in ArcSlider.js was done to get a 'valueAngle' for the 'min' value to have as the default if 'min' is greater or equal to 'max', the same was done to have a minimum 'value' in ArcSliderBehaviorDecorator.js when 'min' is greater or equal to 'max'
- Added componentDidUpdate in ArcSliderBehaviorDecorator.js to set the value to be in range if it is bigger than 'max' or lower than 'min'
- Incremented 'max' in ArcSlider/utils.js to not have division by 0 if 'value', 'min' and 'max' are equal

### Additional Considerations

### Links
PLAT-123306

### Comments


daniel-stoian-lgp (approved):

ArcSlider/ArcSliderBehaviorDecorator.js
```diff
@@ -145,6 +145,16 @@ const ArcSliderBehaviorDecorator = hoc((config, Wrapped) => {
 			};
 		}
 
+		componentDidUpdate (prevProps) {
+			if ( this.props.max !== prevProps.max || this.props.min !== prevProps.min) {
+				if (this.state.value && this.state.value >= this.props.max) {
+					this.handleChange(null, this.props.max);
+				} else if ( this.state.value && this.state.value <= this.props.min) {
+					this.handleChange(null, this.props.min);
```
jeonghee27, line 153: how about use clamp function?

import {clamp} from '@enact/core/util';

tests/screenshot/apps/components/TemperatureControl.js
```diff
@@ -1,7 +1,17 @@
...
+	// Min value equals max value - [PLAT-123306]
+	<TemperatureControl
+		max={30}
+		min={30}
+	/>,
+	// Min value > max value - [PLAT-123306]
+	<TemperatureControl
+		max={30}
+		min={50}
+	/>
```
jeonghee27, line 14: Please add normal cases also.

<TemperatureControl
		max={50}
		min={30}
	/>

ArcSlider/ArcSliderBehaviorDecorator.js
```diff
@@ -180,7 +190,12 @@ const ArcSliderBehaviorDecorator = hoc((config, Wrapped) => {
 
 			const angle = positionToAngle(coordsInSvg, radius * 2 - strokeWidth);
 			// get the value based on the angle, min and max
-			let value = angleToValue(angle, min, max, startAngle, endAngle);
+			let value;
+			if (max <= min) {
+				value = angleToValue(angle, min, min, startAngle, endAngle);
+			} else {
+				value = angleToValue(angle, min, max, startAngle, endAngle);
+			}
```
jeonghee27, line 198:
```suggestion
			const value = angleToValue(angle, min, Math.max(min, max), startAngle, endAngle);
```
How about it? There may be other good alternatives.

ArcSlider/ArcSlider.js
```diff
@@ -210,7 +210,12 @@ const ArcSliderBase = kind({
 
 	render: ({'aria-valuetext': ariaValuetext, backgroundColor, componentRef, disabled, endAngle, foregroundColor, isFocused, max, min, radius, size, slotCenter, startAngle, strokeWidth, value, ...rest}, context) => {
 		const {accent: accentColor} = context || {};
-		const valueAngle = valueToAngle(value, min, max, startAngle, endAngle);
+		let valueAngle;
+		if (max <= min) {
+			valueAngle = valueToAngle(min, min, min, startAngle, endAngle);
+		} else {
+			valueAngle = valueToAngle(value, min, max, startAngle, endAngle);
+		}
```
jeonghee27, line 218: same with ArcSliderBehaviorDecorator.js

jeonghee27:
There is tavis fail. please check it.

ArcSlider/ArcSliderBehaviorDecorator.js
paul-beldean-lgp: good idea

ArcSlider/ArcSliderBehaviorDecorator.js
paul-beldean-lgp: yes, it's better this way, but I will need to use "let" because the value is updated after this

jeonghee27 (approved):
LGTM

PR #507, opened 2021-02-03
PLAT-133689: Preserve LabeledIconButton behavior and look when missing props

stanca-pop-lgp:

### Checklist

### Resolution
Updated styling to preserve LabeledIconButton behavior and look when missing props. Added ellipsis when label is too long and set max-width values depending on the button size and label positioning.

### Links
PLAT-133689

### Comments


LabeledIconButton/LabeledIconButton.module.less
```diff
@@ -15,8 +15,31 @@
...
+		&.small {
+			max-width: calc(291px + @agate-button-small-height);
+		}
+
+		&.large {
+			max-width: calc(291px + @agate-button-height);
+		}
+
+		&.huge {
+			max-width: calc(291px + @agate-button-huge-height);
```
adrian-cocoara-lgp, line 41: 291 + 250 = 541 is not divisible by 3. Maybe change this as: calc(290px + @agate-button-huge-height)

adrian-cocoara-lgp:
I think screenshot tests should be updated with different sizes and situations with children and no children

LabeledIconButton/LabeledIconButton.module.less
stanca-pop-lgp: I will adjust it in Button style. Thank you for noticing.

jeonghee27:
Did you check screenshot result ?
the PR request format was changed.

### Checklist

* [ ] I have read and understand the [contribution guide](http://enactjs.com/docs/developer-guide/contributing/)
* [ ] Documentation was verified or is not changed
* [ ] UI test was passed or is not needed
* [ ] Screenshot test was verified or is not needed

Please check it.

LabeledIconButton/LabeledIconButton.module.less
```diff
@@ -15,8 +15,31 @@
...
+	.label {
+		text-overflow: ellipsis;
+		overflow: hidden;
+		width: 100%;
+		white-space: nowrap;
+	}
+
 	.applySkins({
-		max-width: min-content;
+		max-width: 291px;
```
jeonghee27, line 26: Please make less variable for "291px"

LabeledIconButton/LabeledIconButton.module.less
```diff
@@ -15,8 +15,31 @@
 		/* Public Class Names */
 	}
 
+	.label {
+		text-overflow: ellipsis;
```
jeonghee27, line 19: Could you find a way to show 2 line and apply ellipsis??
It seems impossible, but I'm not sure.

stanca-pop-lgp: I used the "line-clamp" property, and it seems to work only if I add a third word/character to the label. Meaning, I added a " !" at the end of the label in the storybook so that the ellipsis is applied. If I remove the " !" or just stick it to "Hello LabeledIconButton!" so that it doesn't count it as a different word, the label is just cut off, with no ellipsis dots. 
Do you think we should leave it like this or revert to the solution above?
Thanks!

jeonghee27: I think the previous one was better in terms of completeness. And as I talked to the GUI person, it seems to be changed to one line in v2.5. Let's go back to one line.

jeonghee27 (approved):
LGTM.

LabeledIconButton/LabeledIconButton.js
```diff
@@ -106,6 +115,9 @@ const LabeledIconButtonBase = kind({
 		 */
 		selected: PropTypes.bool,
 
+		// forwarded from Button
```
MikyungKim, line 118: Please update jsdoc properly even though it's forwarded.

LabeledIconButton/LabeledIconButton.js
```diff
@@ -94,6 +94,15 @@ const LabeledIconButtonBase = kind({
...
 		iconOnly: PropTypes.bool,
 
+		/**
+		 * The position of the label relative to icon button.
+		 *
+		 * @type {('above'|'after'|'before'|'below')}
+		 * @default 'below'
+		 * @public
+		 */
+		labelPosition: PropTypes.oneOf(['above', 'after', 'before', 'below', 'left', 'right']),
```
MikyungKim, line 104: Doesn't the JSDoc need to be the same?

LabeledIconButton/LabeledIconButton.js
```diff
@@ -94,6 +94,15 @@ const LabeledIconButtonBase = kind({
...
+
```
MikyungKim, line 105: Please write JSDoc for `backgroundOpacity` too.

MikyungKim (changes requested):
Please find my comments. Don't we need a changelog for this?

MikyungKim (approved):
:shipit:

PR #514, opened 2021-02-15
PLAT-129636: Apply max-width in Arc family components

paul-beldean-lgp:
### Checklist

slotCenter content was too big and overflowed the arc component

### Resolution
Added Marquee and max-width to make it display correctly

### Additional Considerations

### Links
PLAT-129636

### Comments


styles/variables-base.less
```diff
@@ -225,6 +227,7 @@
 // ---------------------------------------
 @agate-fanspeedcontrol-icon-font-size:   inherit;
 @agate-fanspeedcontrol-icon-height:      inherit;
+@agate-fanspeedcontrol-icon-line-height: 2rem;
```
daniel-stoian-lgp, line 230: use pixel values in the code. Agate has a resolution independence system that will adapt this value to the screen size.

styles/variables-base.less
```diff
@@ -706,3 +709,4 @@
 @agate-winddirectioncontrol-icon-height: 180px;
 @agate-winddirectioncontrol-icon-line-height: inherit;
 @agate-winddirectioncontrol-icon-width: inherit;
+@agate-winddirectioncontrol-icon-margin: 0px;
```
daniel-stoian-lgp, line 712: 0 values don't need px

daniel-stoian-lgp:
This PR needs a changelog entry

daniel-stoian-lgp (approved):

MikyungKim (changes requested):
I agree to add max-width to `slotCenter`, but for Marquee, I hope we could fix qa-a11y sample.

MikyungKim (approved):
👏 LGTM!

seunghoh (approved):
LGTM 👍 
The further consideration(even not related to this PR): This is A11y sample but there's no way to read out `slotCenter`. Please consider how can we support to read `slotCenter` with 5-way key when the ArcPicker, ArcSlider get focused.

PR #517, opened 2021-02-17
PLAT-135587: Updated Silicon skin styles for Button

daniel-stoian-lgp:
### Checklist

Button component needs to be updated to the latest design changes

### Resolution
- Added styling for Button based on the latest UX design for Silicon
- Because of the new styling for iconOnly Button, other components were affected, so new styles were added to restore those styles to the current value from the develop branch ( ColorPicker, ColorPicker/SwatchButton, Keypad, IncrementSlider/IncrementSliderButton )
- Because @agate-button-height has changed for silicon, its' usage DateTimePicker, SliderButton and internal/Picker was changed to a static value(the previous value of agate-button-height)
- @silicon-gradient core color has a new value on the UX guide. The change of its' value got many ss-tests to fail, but this is normal
- the AgateSilicon ss-tests for other components that use Button or Icon are failing because those components use Button or Icon. They will be fixed when those component`s design will be updated.
- update .active() mixin to include also pressed state ( active state of Button was not triggered on select Enter key until now)
- added new screenshot tests for Button

### Additional Considerations

### Links
PLAT-135587

### Comments


stanca-pop-lgp (changes requested):
Please check out the styling for selected. It does not seem to apply style from Zeplin in Storybook when "selected" knob is checked.

styles/variables-silicon.less
```diff
@@ -37,17 +37,39 @@
...
 @agate-button-line-height: 1.17em;
+@agate-button-max-width: 420px;
+@agate-button-min-width: 180px;
+@agate-button-icon-margin-end: 9px;
+@agate-button-icon-margin-start: 0;
 @agate-button-huge-border-width: @agate-button-border-width;
+@agate-button-huge-font-size: 33px;
+@agate-button-huge-h-padding: @agate-button-h-padding;
+@agate-button-huge-height: 180px;
+@agate-button-huge-max-width: 300px;
```
stanca-pop-lgp, line 57: This should be 360px.

daniel-stoian-lgp:
> Also the icon's top and bottom margins for all button sizes.

Since the icon is vertically aligned at middle with the text, apllying those padding/margins have no  visual effect

stanca-pop-lgp (approved):
LGTM

tests/screenshot/apps/components/Button.js
```diff
@@ -1,10 +1,161 @@
...
+	<Button icon="arrowharpoonright" iconFlip="vertical">click me</Button>,
+	<Button icon="arrowharpoonright" iconFlip="both">click me</Button>,
+	<Button icon="arrowharpoonright" iconFlip="auto">click me</Button>,
+
+
+	// *************************************************************
+	// Tallglyph validation
+	// locale = 'vi-VN'
+	// *************************************************************
+	...withConfig({locale: 'vi-VN'}, [
```
jeonghee27, line 67: Could you add focused case also?

...withConfig({focus: true}, [
...
]

styles/colors-silicon-day.less
```diff
@@ -80,21 +80,28 @@
...
-@agate-button-bg-color:       @agate-white;
+@agate-button-bg-color:       @silicon-gray;
 @agate-button-bg-image:       none;
-@agate-button-border-color:   fade(@agate-black, 50%);
+@agate-button-border-color:   transparent;
 @agate-button-shadow:         none;
 @agate-button-active-color:   @agate-button-color;
-@agate-button-active-bg-color: @silicon-light-gray;
+@agate-button-active-bg-color: @silicon-gray;
+@agate-button-disabled-bg-color: fade(@agate-button-bg-color, 50%);
```
jeonghee27, line 89: 50%? I saw 30%

Button/Button.module.less
```diff
@@ -42,14 +42,24 @@
...
+			padding: @agate-button-client-padding;
 			color: @agate-button-color;
 
 			.icon {
 				.margin-start-end(@agate-button-icon-margin-start, @agate-button-icon-margin-end);
 			}
 		}
 
+		.disabled({
+			.bg {
```
jeonghee27, line 54: For disabled style, opacity 0.6 was applied before.
This impede consistency depending on the background color.
So, GUI person said that some fixed color will apply for disabled. Could you remove the opacity style that was previously applied only to silicon?
CC : @adrian-cocoara-lgp @stanca-pop-lgp @paul-beldean-lgp @0x64 @MikyungKim

daniel-stoian-lgp: Alright. This was actually the cause of why I used 50% opacity for colors instead of the 30% from the GUI guide.

daniel-stoian-lgp: @jeonghee27  Actually, is it ok to set @agate-disabled-opacity to 1? That will affect all components. Or maybe I misunderstood which previosuly opacity of 0.6 you meant.

jeonghee27: Oh, I didn't say it clearly. First of all, I wanted to remove it only from the silicone button.
And, it still looks opacity applied. It doesn't look like #444444.
image

jeonghee27:
The press effect looks different from the GUI document. Could you please check again?

Button/Button.module.less
daniel-stoian-lgp: I don't know what happened, but apparently, the @agate-disabled-opacity: 1 setting was not pushed. I fixed now the opacity only for Silicon Button

daniel-stoian-lgp:
> The (focus+press) effect looks different from the GUI document. Could you please check again?

Fixed it :)

jeonghee27:
Silicon works well, but there is regression in the press effect and disable in Gallium. please check :(

jeonghee27:
image
Regression is in the press effect and disable in Gallium. :(

After correction, please review again for the operation. @stanca-pop-lgp

daniel-stoian-lgp:
>
> Regression is in the press effect and disable in Gallium. :(
> 
> After correction, please review again for the operation. @stanca-pop-lgp

@jeonghee I pushed a fix.
All ss-tests passed. I had to introduce an extra .vendor-opacity() styling for LabeledIcoButton because the new disabled opacity for Button had a bigger specificity and introduced a regression for a disabled LabeledIconButton.

stanca-pop-lgp:
It seems that the text is not vertically aligned in the middle for the smallest size.

stanca-pop-lgp:
Please also adjust styling for joinedPosition when button is smallest.

daniel-stoian-lgp:
> It seems that the text is not vertically aligned in the middle for the smallest size. Could you check it out?

For me it looks ok

image

stanca-pop-lgp:
> > It seems that the text is not vertically aligned in the middle for the smallest size. Could you check it out?
> 
> For me it looks ok
> 
> image

image

daniel-stoian-lgp:
> > > It seems that the text is not vertically aligned in the middle for the smallest size. Could you check it out?
> > 
> > 
> > For me it looks ok
> > image
> 
> image
> 
> image

I can't reproduce it on my computer

image

stanca-pop-lgp:
> > > > It seems that the text is not vertically aligned in the middle for the smallest size. Could you check it out?
> > > 
> > > 
> > > For me it looks ok
> > > image
> > 
> > 
> > image
> > image
> 
> I can't reproduce it on my computer
> 
> image

Try applying the same value to line-height as to height.

styles/variables-silicon.less
```diff
@@ -32,22 +32,42 @@
...
+@agate-button-huge-font-size: 33px;
+@agate-button-huge-h-padding: @agate-button-h-padding;
+@agate-button-huge-height: 180px;
+@agate-button-huge-max-width: 360px;
+@agate-button-huge-min-width: 180px;
 @agate-button-small-border-radius: (@agate-button-small-height / 2);
-@agate-button-small-font-size: 28px;
-@agate-button-smallest-height: 39px;
-@agate-button-smallest-font-size: 18px;
+@agate-button-small-font-size: 25px;
```
stanca-pop-lgp: Multiple of 3. (also design says 26)

jeonghee27 (approved):
LGTM

PR #519, opened 2021-02-17
PLAT-135589: Update BodyText

stanca-pop-lgp:

### Checklist

### Resolution
Updated BodyText according to the new design requirements. Updated styling and wrapped component in Scroller. 

### Links
PLAT-135589

### Comments


BodyText/BodyText.js
```diff
@@ -18,6 +18,7 @@ import compose from 'ramda/src/compose';
 import React from 'react';
 
 import {MarqueeDecorator} from '@enact/ui/Marquee';
```
daniel-stoian-lgp, line 20: move import to first section

BodyText/BodyText.js
```diff
@@ -118,11 +129,13 @@ const BodyTextBase = kind({
 			);
 		}
 		return (
-			<UiBodyText
-				{...rest}
-				centered={centered}
-				css={css}
-			/>
+			<Scroller verticalScrollbar="visible" focusableScrollbar={focusableScrollbar}>
```
daniel-stoian-lgp, line 132: the scroller should not be visible all the time

stanca-pop-lgp: After refactoring the Scroller it will not be, as it won't have the buttons and you will only see the scrollTrack when content should be scrolled. If I change to verticalScrollbar="auto", the scrollbar won't take up its designated space.

daniel-stoian-lgp: Alright. Thanks!

daniel-stoian-lgp (approved):

jeonghee27:
Please  resolve the conflict and merge develop again. 
The number of file changes is strange.

BodyText/BodyText.js
jeonghee27: I don't think we should wrap the BodyText with Scroller by default. if needed, The app will put in a scroller .
Everything in the GUI doesn't have to be provided by the component  side. Agate provides just the parts that can assembles.

samples/sampler/stories/default/BodyText.js
```diff
@@ -18,6 +18,7 @@ storiesOf('Agate', module)
 		() => (
 			<BodyText
 				centered={boolean('centered', Config)}
+				disabled={boolean('disabled', Config)}
```
jeonghee27, line 21: Why this prop is needed for BodyText?

jeonghee27: I think we can remove this. Please check it when merge.

jeonghee27 (commented):
LGTM except for `disabled` prop. we need to check it is needed.

samples/sampler/stories/default/BodyText.js
stanca-pop-lgp: You are right, it is not in the GUI. Thanks!

PR #531, opened 2021-03-03
PLAT-136487: Update Silicon styles for RadioItem

adrian-cocoara-lgp:
### Checklist

RadioItem component needs to be updated to the latest design changes

### Resolution
- Added styling for RadioItemItem based on the latest UX design for Silicon

### Additional Considerations

### Links
PLAT-136772

### Comments


styles/colors-base.less
```diff
@@ -226,6 +226,7 @@
 @agate-item-selected-border-color:        @agate-item-border-color;
 @agate-item-selected-color:               @agate-item-color;
 @agate-item-selected-label-color:         inherit;
+@agate-item-focus-bg-image:               none;
```
stanca-pop-lgp, line 230: Please add min-width and max-width properties as well.

samples/sampler/stories/default/RadioItem.js
```diff
@@ -24,6 +24,7 @@ storiesOf('Agate', module)
 					disabled={boolean('disabled', Config)}
 					icon={select('icon', ['', ...iconNames], Config)}
 					onToggle={action('onToggle')}
+					inline={boolean('inline', Config)}
```
stanca-pop-lgp, line 27: Please add the "selected" knob and sort props alphabetically.

stanca-pop-lgp (approved):

daniel-stoian-lgp (commented):
2 minor comments

styles/colors-base.less
```diff
@@ -340,12 +340,14 @@
 // RadioItem
 // ---------------------------------------
 @agate-radioitem-focus-color: inherit;
+@agate-radioitem-focus-border-color: transparent;
```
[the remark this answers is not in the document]
daniel-stoian-lgp: I retract this

RadioItem/RadioItem.module.less
```diff
@@ -30,6 +46,8 @@
 		});
 
 		&.selected {
+			border-color: @agate-radioitem-selected-border-color;
```
daniel-stoian-lgp, line 49: is this needed? Doesn't it come from Item?

daniel-stoian-lgp (approved):

RadioItem/RadioItem.module.less
```diff
@@ -5,11 +5,9 @@
 
 // Public
 .radioItem {
-	.icon {
-		margin: 0;
-	}
-
 	.applySkins({
+		height: @agate-radioitem-height;
```
MikyungKim, line 9: 🤔 I think I see this variation a lot...
The default size of RadioItem of silicon seems "small" right?
I think I guess just pass `size="small"` to the Item when silicon skin is applied instead of overriding the size alone.

adrian-cocoara-lgp: Sounds good to me. Thanks for the tip

RadioItem/RadioItem.js
```diff
@@ -70,6 +78,14 @@ const RadioItemBase = kind({
 		 */
 		selected: PropTypes.bool,
 
+		/**
+		 * The current skin for this component.
+		 *
+		 * @type {String}
+		 * @pub
```
MikyungKim, line 85:
```suggestion
		 * @private
```

RadioItem/RadioItem.js
```diff
@@ -91,16 +107,19 @@ const RadioItemBase = kind({
...
+	render: ({children, css, icon, selected, skin, slotBefore, ...rest}) => {
+		const itemSize = skin === 'silicon' ? 'small' : 'large';
+
 		return (
 			<Item
 				aria-checked={selected}
 				role="checkbox"
 				{...rest}
 				css={css}
+				size={itemSize}
```
MikyungKim, line 122: Let's not make a constant here.
```suggestion
				size={skin === 'silicon' ? 'small' : 'large'}
```

RadioItem/RadioItem.module.less
```diff
@@ -18,10 +14,28 @@
...
+		}
+
+		&.inline {
+			&.item {
+				padding: @agate-radioitem-inline-padding;
+			}
+		}
+
+		.slotBefore {
+			padding: @agate-radioitem-slotbefore-padding;
```
MikyungKim, line 30: Do we have a guid for this?
image

RadioItem/RadioItem.module.less
```diff
@@ -18,10 +14,28 @@
 			border-radius: @agate-radioitem-icon-border-radius;
 			color: transparent;
 			opacity: @agate-radioitem-icon-opacity;
+			margin: @agate-radioitem-icon-margin;
+			display: flex;
+			align-items: center;
+			justify-content: center;
```
MikyungKim, line 20: Why did you introduce this?
I think we could get the same result with `margin: 15px` here without display: flex.

RadioItem/RadioItem.module.less
adrian-cocoara-lgp: I added padding here to follow this requirement:
padding

..but since it breaks layout when `slotBefore` icon is used, I will remove it.
I will add margin to icon to follow the above requirement, and this way the `slotBefore` won't be affected

RadioItem/RadioItem.module.less
adrian-cocoara-lgp: 👍

RadioItem/RadioItem.module.less
```diff
@@ -18,10 +14,24 @@
...
+			&.item {
+				padding: @agate-radioitem-inline-padding;
+			}
+		}
+
+		.slotBefore {
+			display: flex;
+			align-items: center;
+			margin-inline-end: @agate-radioitem-slotbefore-margin-end;
 		}
```
MikyungKim, line 30: I think we don't need these.

RadioItem/RadioItem.module.less
```diff
@@ -18,10 +14,24 @@
...
+
+		&.inline {
+			&.item {
+				padding: @agate-radioitem-inline-padding;
+			}
+		}
+
+		.slotBefore {
+			display: flex;
+			align-items: center;
```
MikyungKim, line 28: I think we don't need these.

adrian-cocoara-lgp: You're right, we don't. I deleted it.

MikyungKim (approved):
LGTM

jeonghee27:
There is no test case for inline.

RadioItem/RadioItem.module.less
```diff
@@ -18,10 +14,22 @@
 			border-radius: @agate-radioitem-icon-border-radius;
 			color: transparent;
 			opacity: @agate-radioitem-icon-opacity;
+			margin: @agate-radioitem-icon-margin;
+		}
+
+		&.inline {
+			&.item {
+				padding: @agate-radioitem-inline-padding;
```
jeonghee27, line 22: I don't know why the padding should diverge depending on "inline". 
Did you add it because there is only an inline case in the GUI?
I think it's better to be unified.

adrian-cocoara-lgp: Okay. We can remove this

jeonghee27: OK, Let's unify.

PR #543, opened 2021-03-12
PLAT-137553: Update FanSpeedControl

paul-beldean-lgp:
### Checklist

FanSpeedControl component needs to be updated to the latest design changes

### Resolution
Added styling for FanSpeedControl based on the latest UX design for Silicon

### Additional Considerations

### Links
PLAT-137553

### Comments


daniel-stoian-lgp:
Check the other skins please. You can use the same values as for Gallium :)

daniel-stoian-lgp:
add changelog and update the PR description

stanca-pop-lgp:
Please link QWA-7319 test case and update PR description.

daniel-stoian-lgp (approved):

stanca-pop-lgp:
Shouldn't the colours be inverted? 
image
Zeplin: 
image

stanca-pop-lgp (approved):
LGTM

jeonghee27:
Could you add more screenshot test to cover all GUI cases?

FanSpeedControl/FanSpeedControl.js
```diff
@@ -112,19 +119,25 @@ const FanSpeedControlBase = kind({
...
+		delete rest.skinVariants;
+
 		return (
 			<div {...rest}>
 				<ArcPicker
 					{...ariaProps}
+					backgroundColor={backgroundColor}
 					disabled={disabled}
 					endAngle={312}
+					foregroundColor={foregroundColor}
```
jeonghee27, line 140: Can't we just give the default color in arcPicker without fixing this?
FanSpeedControl and ArcPicker are not the same color?
If you do that, both components will be modified with one code.

styles/variables-silicon.less
```diff
@@ -126,9 +126,12 @@
 
 // FanSpeedControl
 // ---------------------------------------
-@agate-fanspeedcontrol-icon-font-size:   120px;
-@agate-fanspeedcontrol-icon-height:      60px;
-@agate-fanspeedcontrol-value-width:      90px;
+@agate-fanspeedcontrol-icon-font-size: 120px;
+@agate-fanspeedcontrol-icon-height: 60px;
```
jeonghee27, line 130: Same value as variables-base, should we define it again?

FanSpeedControl/FanSpeedControl.js
paul-beldean-lgp: Ok, I reverted the changes from FanSpeedControl.js

jeonghee27: Then, will the ArcPicker be modified in other issues? 
I cannot find issue for ArcPicker.

The color has not been corrected.
https://github.com/enactjs/agate/pull/543#issuecomment-801834279

tests/screenshot/apps/components/FanSpeedControl.js
```diff
@@ -1,7 +1,15 @@
 import FanSpeedControl from '../../../../FanSpeedControl';
 
 const FanSpeedControlTests = [
+	<FanSpeedControl />,
+	<FanSpeedControl min={1} />,
+	<FanSpeedControl max={10} />,
+	<FanSpeedControl value={5} />,
+	<FanSpeedControl icon="fan" min={1} max={10} />,
+	<FanSpeedControl icon="fan" min={1} max={10} value={1} />,
 	<FanSpeedControl min={1} max={10} value={5} />,
```
jeonghee27, line 10: I think non-icon cases are not necessary. (There is not in GUI)
Instead, Please add disabled cases.

FanSpeedControl/FanSpeedControl.js
jeonghee27: I think you missed this comment.

paul-beldean-lgp: Yes, sorry, I missed the comment, I added the default values in ArcPicker. 
Also there is no ticket for ArcPicker because other than the colors there were no other changes requested for it.

jeonghee27:
Night mode is good.
But. Could you check day mode?
image
Although there is no GUI guide for day mode, please make it look appropriately like other components.

jeonghee27:
> @jeonghee27 Should I implement this in the sampler? Because we can't have defaultProp based on component skinVariant

No we want to fix component. why don't you remove defaultProp and set it on "computed" depend on skinVariant?

And we don't like the color value is in js. How about use css variables on colors-silicon-day.less and colors-silicon-night.less?

paul-beldean-lgp:
> No we want to fix component. why don't you remove defaultProp and set it on "computed" depend on skinVariant?
> 
> And we don't like the specific color value is in js. How about use css variables on colors-silicon-day.less and colors-silicon-night.less?
> I'm not sure this works or not.

@jeonghee27 
Regarding setting the color based on skinVariants in computed block, this can be done if no color is given by the consumer.

Regarding using css variables instead of colors in the js file, this would lead to a great architecture change for ArcPicker/Arc.
When we created these components, we got the recommendation from the SWP team that Arc and ArcPicker/ArcSlider should be as flexible as possible, with the consumer being able to choose the background and foreground colors. That's why those are props.
If we change to using css variables, then these 2 props will have to be removed and the color change flexibility will be removed.
This change will need a broader  number of opinions from the team and would also need another user story to be solved in.

ArcPicker/ArcPicker.js
```diff
@@ -159,9 +159,9 @@ const ArcPickerBase = kind({
 	},
 
 	defaultProps: {
-		backgroundColor: '#eeeeee',
+		backgroundColor: '#444444',
 		endAngle: 310,
-		foregroundColor: '#444444',
+		foregroundColor: '#ffffff',
```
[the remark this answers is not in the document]
paul-beldean-lgp: That is a good solution, I didn't think of doing it like that as I thought the prop should have a value by default.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ The following is a curated list of changes in the Enact agate module, newest cha
 
 - `agate/ContextualPopup` and `agate/ContextualPopupDecorator` prop `css` to support customizations
 - `agate/Dropdown` prop `width` to support multiple widths
+- `agate/FanSpeedControl` style to match latest design for Silicon skin
```
jeonghee27, line 19: Please move to [unreleased] fixed

paul-beldean-lgp: Done

jeonghee27 (approved):
LGTM

FanSpeedControl/FanSpeedControl.module.less
```diff
@@ -22,11 +22,14 @@
 			height: @agate-fanspeedcontrol-icon-height;
 			line-height: @agate-fanspeedcontrol-icon-height;
 			margin: 0;
+			padding-left: @agate-fanspeedcontrol-icon-padding-left;
```
MikyungKim, line 25: Would you please consider RTL case?

FanSpeedControl/FanSpeedControl.module.less
```diff
@@ -22,11 +22,14 @@
...
 			width: @agate-fanspeedcontrol-icon-height;
 		}
 
 		.fanValue {
 			width: @agate-fanspeedcontrol-value-width;
+			padding-left: @agate-fanspeedcontrol-value-padding-left;
```
MikyungKim, line 31: Would you please consider RTL case?

MikyungKim (changes requested):
UI tests are broken with this change. please check.

paul-beldean-lgp:
Updated the UI tests as the default ArcPicker colors have changed.

MikyungKim (approved):
:shipit:

PR #544, opened 2021-03-15
PLAT-137580: Update WindDirectionControl

paul-beldean-lgp:
### Checklist

WindDirectionControl component needs to be updated to the latest design changes

### Resolution
Added styling for WindDirectionControl based on the latest UX design for Silicon

### Additional Considerations

### Links
PLAT-137580

### Comments


daniel-stoian-lgp:
add changelog and update the PR description

stanca-pop-lgp:
Also link QWA-7292 test case and update PR description.

MikyungKim (changes requested):
When `night` mode, we can't see the icon like below.
image

styles/variables-silicon.less
```diff
@@ -470,7 +470,7 @@
...
 // WindDirectionControl
 // ---------------------------------------
-@agate-winddirectioncontrol-icon-font-size: 210px;
-@agate-winddirectioncontrol-icon-height: 210px;
-@agate-winddirectioncontrol-icon-line-height: 210px;
-@agate-winddirectioncontrol-icon-width: 210px;
+@agate-winddirectioncontrol-icon-font-size: 240px;
+@agate-winddirectioncontrol-icon-height: 240px;
+@agate-winddirectioncontrol-icon-line-height: 240px;
+@agate-winddirectioncontrol-icon-width: 240px;
```
MikyungKim, line 476: Where comes these numbers from?

paul-beldean-lgp:
I'll keep this ticket on my side until PLAT-137553(https://github.com/enactjs/agate/pull/543) is merged, as the solution for the color bug is in that ticket.

styles/variables-silicon.less
paul-beldean-lgp: It's double the size of the icon element in the GUI, so that the icon looks the same size in the sampler as it does in the GUI.

paul-beldean-lgp:
Now that https://github.com/enactjs/agate/pull/543 was merged, the bug for night mode is solved.

styles/variables-silicon.less
```diff
@@ -484,7 +484,7 @@
...
```
MikyungKim, line 490: I think the size of the icon and line-height are 120 px and the margin should be 0.
image

paul-beldean-lgp: @MikyungKim 
If I use a font-size of 120px and a line height of 120px it looks like this:
image
The icon is very small and because of the line height it is no longer centered.

With the font-size of 240px and line height of 240px it looks like this:
image

And this is the comparison of the 240px size and the GUI
image

MikyungKim: I didn't mean to change `font-size`. I meant the size and line-height to be 120px. Please re-check.

MikyungKim: That's why I dragged those three lines(488 to 490) to comment on this.

paul-beldean-lgp: I solved the issues and removed the margin, but I could only remove he margin by adding it inside of 'applySkins' otherwise it would not apply.

MikyungKim (approved):
LGTM

PR #560, opened 2021-04-16
PLAT-139507: Fixed Layout bug for inline item on Gallium skin

stanca-pop-lgp:

### Checklist

### Resolution
Fixed Layout bug for inline item with label and labelPosition "after" or "before" on Gallium skin to not change item width on hover. Also updated screenshot and ui tests.

### Links
PLAT-139507

### Comments


adrian-cocoara-lgp (approved):
LGTM

styles/variables-base.less
```diff
@@ -415,18 +415,18 @@
 
 // Item
 // ---------------------------------------
-@agate-item-border-width:             inherit;
+@agate-item-border-width:             0;
 @agate-item-content-width:            60%;
-@agate-item-centered-inline-max-width: @agate-item-inline-max-width;
+@agate-item-centered-inline-max-width: @agate-item-inline-min-width;
```
jeonghee27, line 420: Why we should change this?

jeonghee27 (approved):
LGTM. Please check @agate-item-centered-inline-max-width when merge

styles/variables-base.less
stanca-pop-lgp: Yes, also check "centered" knob. It is only valid for centered. If I undo this change, when hovered, the item becomes wider. 
For inline item with centered=false the max-width is indeed @agate-item-inline-max-width.

PR #572, opened 2021-04-29
PLAT-141501: Heading - line under title has wrong position

adrian-cocoara-lgp:
### Checklist

Underline for heading when `size="title"` has a wrong position for Carbon, Cobalt, Copper, Electro, Titanium skins

### Resolution
Adjusted padding

### Additional Considerations

### Links
PLAT-141501

### Comments


styles/variables-carbon.less
```diff
@@ -36,6 +36,7 @@
 @agate-heading-title-decoration-border-radius: 3px;
 @agate-heading-title-decoration-opacity: 0.8;
 @agate-heading-title-decoration-transform: skew(-10deg);
+@agate-heading-title-padding: 0 51px 21px 51px;
```
daniel-stoian-lgp, line 39: can you use this value in variables-base? 
An remove the variables in the files specific for skins, since they all use the same value

daniel-stoian-lgp (approved):
LGTM

styles/variables-base.less
```diff
@@ -306,7 +306,7 @@
 @agate-heading-font-ligatures: @agate-base-font-ligatures;
 @agate-heading-title-height: auto;
 @agate-heading-title-line-height: inherit;
-@agate-heading-title-padding: @agate-heading-padding 50px;
+@agate-heading-title-padding: 0 51px 21px 51px;
```
jeonghee27, line 309: I think we should add margin-padding only when showline.
and Please add screenshot test for showline.

adrian-cocoara-lgp: I don't see why I should delete the padding-bottom here, sorry if I don't understand correctly. It would look like this:
heading-padding

_It is also a bug that showline =false is not woking on the sampler._
Could you give more details? I see no problem with `showline` prop

adrian-cocoara-lgp: Added more ss tests

jeonghee27: It's my misunderstand.
I found that the center line and showline have nothing to do with it.

jeonghee27 (approved):
LGTM

PR #604, opened 2021-06-15
WRO-6459: Updated Silicon styles for Picker and RangePicker

daniel-stoian-lgp:
### Checklist

- Updated silicon styles for Picker/RangePicker
- Fixed ss tests for Date and Time Picker - they were falsey passing because the value prop was different at each run, so there was no reference to compare to
- You may notice that some spacing(padding/margin) for Picker item from the design is not there and the height is different from the design.The padding was added to the container width/height to avoid ViewManager transition effect issues. Since those containers have transparent bg-color, the visual result is the same and the viewManager transition effect is not disturbed
- The layout with 5 items was kept for silicon skin. The UX guide is confusing because it shows picker with 3 and 5 values on the same screen. If we should switch to 3 items, let us know.
- `backgroundFocusClassName` was added in computed block in order for it to inherit the class name from `DateComponentPicker` component
- The z-indexes were set in order to obtain the following layout `active background --> focus background --> text`. There was an option to just set z-index:-1 for the .active background, but it would make the ss tests to fail because the active bakground would be covered by the ss tests view background. Any other working option would be welcomed

### Resolution

### Additional Considerations

### Links
WRO-6459

### Comments


adrian-cocoara-lgp (approved):
LGTM

styles/variables-silicon.less
```diff
@@ -345,9 +345,14 @@
 
 // Picker
 // ---------------------------------------
-@agate-picker-font-size: 27px;
+@agate-picker-font-size: 33px;
 @agate-picker-font-weight: normal;
```
stanca-pop-lgp, line 349: It seems that in the design the font-weight is bold.

styles/variables-silicon.less
```diff
@@ -345,9 +345,14 @@
...
 @agate-picker-active-font-weight: bold;
+@agate-picker-active-height: 99px;
+@agate-picker-horizontal-active-width: 201px;
+@agate-picker-horizontal-item-height: 160px;
+@agate-picker-horizontal-item-width: 249px;
+@agate-picker-horizontal-padding: 9px 6px;
 @agate-picker-item-height: 75px;
```
stanca-pop-lgp, line 356: Shouldn't this be 60px?

daniel-stoian-lgp: check PR description

styles/variables-silicon.less
daniel-stoian-lgp: fixed

stanca-pop-lgp:
Please add style for active value in disabled state.

internal/Picker/Picker.module.less
```diff
@@ -73,33 +90,83 @@
...
+				z-index: 2;
+			}
+
+			&::before {
+				content: "";
+				background-color: @agate-picker-active-bg-color;
+				background-image: @agate-picker-active-bg-image;
+				position: absolute;
+				height: @agate-picker-active-height;
+				width: calc(100% - 2* @agate-picker-padding);
```
stanca-pop-lgp, line 121: Please leave space after 2 😃

styles/variables-silicon.less
```diff
@@ -345,9 +345,14 @@
...
 // Picker
 // ---------------------------------------
-@agate-picker-font-size: 27px;
-@agate-picker-font-weight: normal;
+@agate-picker-font-size: 33px;
+@agate-picker-font-weight: bold;
 @agate-picker-active-font-weight: bold;
+@agate-picker-active-height: 99px;
+@agate-picker-horizontal-active-width: 201px;
+@agate-picker-horizontal-item-height: 160px;
```
stanca-pop-lgp, line 353: Please add multiple of 3 -> 159px

stanca-pop-lgp (approved):
LGTM

seunghoh (changes requested):
Let's revisit later when we are requested.
I just check request changes not for mering accidently

adrian-cocoara-lgp (approved):
LGTM 👍

seunghoh (approved):

alexandrumorariu (approved):
LGTM

PR #607, opened 2021-06-28
PLAT-122723: Fix VirtualList 5-way navigation between scroll buttons when `focusableScrollbars`

MikyungKim:

### Checklist

5-way `down` from `up` scrollbar button goes to a list item.

### Resolution
`createRef` returns a new ref on every render so it should be `useRef`.
Removed [] dependency since it needs to be updated when props changed.

### Additional Considerations

### Links
PLAT-122723


jeonghee27 (approved):
LGTM

PR #620, opened 2021-08-10
WRN-792: Fix testcases for latest chrome driver 

jeonghee27:
### Checklist

When we use latest chrome driver not 2.44, some test cases are failed due to api change.

1.
element not interactable: element has zero size

2.
move target out of bounds
  (Session info: headless chrome=87.0.4280.141)
[chrome 87.0.4280.141 linux #0-15] move target out of bounds

3.
browser.positionClick is not a function
[chrome 87.0.4280.141 linux #0-15] TypeError: browser.positionClick is not a function

### Resolution
- positionClick() is replaced with click();
- Prevent to move target out of bound
- Interact (eg. click) with "#floatLayer>div"instead of "#floatLayer"because the size is zero. 

### Additional Considerations

### Links
WRN-792
Jenkins pass :

### DIFF
--- a/tests/ui/specs/Popup/PopupPage.js
+++ b/tests/ui/specs/Popup/PopupPage.js
@@ -98,7 +98,7 @@ class PopupPage extends Page {
 	}
 
 	clickPopupFloatLayer () {
-		$('#floatLayer').click();
+		$('#floatLayer>div').click();
 	}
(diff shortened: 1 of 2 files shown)


MikyungKim (approved):
LGTM

PR #650, opened 2022-05-04
ENACT-131: Fixed Spotlight remains on ArcPicker after tap

stanca-pop-lgp:

### Checklist

Spotlight remains on ArcPicker after tap.

### Resolution
- Added a setTimeout to handle Spotlight on ArcPicker for touch devices.

### Additional Considerations

### Links
ENACT-131


stanca-pop-lgp:
Build fails because unit tests in develop branch are still using 'enzyme'. ("Cannot find module 'enzyme'")

daniel-stoian-lgp (commented):
I don't think adding onPointerDown and onPointerUp brings any benefits to the implementation. They seem redundant.
Please test on branch feature/ENACT-131-DS. On Android it seems to fix the problem with less code https://github.com/enactjs/agate/compare/develop...feature/ENACT-131-DS

stanca-pop-lgp:
> I don't think adding onPointerDown and onPointerUp brings any benefits to the implementation. They seem redundant. Please test on branch feature/ENACT-131-DS. On Android it seems to fix the problem with less code [develop...feature/ENACT-131-DS](https://github.com/enactjs/agate/compare/develop...feature/ENACT-131-DS) If this fixes the issue, please also update the changelog with description about the bugfix
> 
> Or are they designed for stylus? I'm not sure this should be a useCase for Agate

It only fixes the "click" part, meaning a quick tap (both finger and pen). If you simulate the hover effect you must keep your finger on the touchscreen for a little longer and release it afterwards (like you would take the mouse off your element). While you keep it on the screen it should be focused. I tried your branch and it doesn't have this behaviour. This is why I added those events so, the way I see it, it is not redundant.

vladut-clapou-lgp (approved):
LGTM

juwonjeong (approved):
LGTM

PR #679, opened 2022-09-22
WRO-12023: Allow double tap behavior on Input

jeonghee27:
### Checklist

Cannot select text via double tap

### Resolution
It was side effect from https://github.com/enactjs/agate/pull/354
We should allow default browser behavior when user touch / mouse-down `input` tag.
So I add condition to `preventDefault`

### Additional Considerations

### Links
WRO-12023
PLAT-94190

### Comments
Enact-DCO-1.0-Signed-off-by: Jeonghee Ahn (jeonghee27.ahn@lge.com)


juwonjeong (approved):
LGTM

PR #712, opened 2022-12-21
WRP-4939: TabGroup - onClick is not supplied to Tab

andreitirlalge:
### Checklist

* [ ] UI test was passed or is not needed
* [ ] Screenshot test was verified or is not needed

Added onSelect function to each Tab inside TabGroup and created a unit test for TabGroup that checks onTabClick.

### Resolution
Each Tab inside TabGroup now have an onSelect function.

### Additional Considerations

### Links
WRP-4939

### Comments
Enact-DCO-1.0-Signed-off-by: Andrei Tirla andrei.tirla@lgepartner.com

### DIFF
--- a/TabGroup/TabGroup.js
+++ b/TabGroup/TabGroup.js
@@ -10,6 +10,7 @@
  * @exports TabGroupDecorator
  */
 
+import handle, {forward, forwardCustom} from '@enact/core/handle';
 import kind from '@enact/core/kind';
 import {cap} from '@enact/core/util';
 import Group from '@enact/ui/Group';
@@ -73,7 +74,7 @@ const TabBase = kind({
 		 * @type {Function}
 		 * @public
 		 */
-		onClick: PropTypes.func,
+		onTabClick: PropTypes.func,
 
 		/**
 		 * Orientation of the tab.
@@ -102,6 +103,13 @@ const TabBase = kind({
 	styles: {
 		css: componentCss,
 		className: 'tab'
+	},
+
+	handlers: {
+		onClick: handle(
+			forward('onClick'),
+			forwardCustom('onTabClick', (ev, {index}) => ({selected: index}))
+		)
 	},
 
 	computed: {
@@ -137,15 +145,16 @@ const TabBase = kind({
 		}
 	},
 
-	render: ({onClick, orientation, style = {}, tabLabel, ...rest}) => {
+	render: ({orientation, style = {}, tabLabel, ...rest}) => {
 		delete rest.labelPosition;
+		delete rest.onTabClick;
 		delete rest.selected;
 
 		if (orientation === 'horizontal') {
 			style.textAlign = 'center';
 		}
 		return (
-			<Cell {...rest} style={style} onClick={onClick}>
+			<Cell {...rest} style={style}>
 				{tabLabel}
 			</Cell>
 		);
@@ -213,6 +222,18 @@ const TabGroupBase = kind({
 		css: PropTypes.object,
 
 		/**
+		 * Called when an item is selected.
+		 *
+		 * The event payload will be an object with the following members:
+		 * * `data` - The value for the option as received in the `children` prop
+		 * * `selected` - Number representing the selected option, 0 indexed
+		 *
+		 * @type {Function}
+		 * @public
+		 */
+		onSelect: PropTypes.func,
+
+		/**
 		 * Orientation of the tabs.
 		 *
 		 * * Values: `'horizontal'`, `'vertical'`
@@ -260,7 +281,7 @@ const TabGroupBase = kind({
 		})
 	},
 
-	render: ({afterTabs, beforeTabs, className, css, labelPosition, orientation, selectedIndex, style, ...rest}) => {
+	render: ({afterTabs, beforeTabs, className, css, labelPosition, onSelect, orientation, selectedIndex, style, ...rest}) => {
 		delete rest.tabPosition;
 		delete rest.tabs;
 
@@ -281,6 +302,7 @@ const TabGroupBase = kind({
 							orientation,
 							shrink: (orientation === 'vertical')
 						}}
+						onSelect={onSelect}
 						orientation={orientation}
 						select="radio"
 						selected={selectedIndex}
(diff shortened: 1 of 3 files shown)


daniel-stoian-lgp:
please add a changelog and update also PR description with that info

daniel-stoian-lgp (approved):

samples/sampler/stories/default/TabGroup.js
```diff
@@ -1,3 +1,4 @@
+import {action} from "@enact/storybook-utils/addons/actions";
 import {mergeComponentMetadata} from '@enact/storybook-utils';
```
stanca-pop-lgp, line 2: Please move this import above.

CHANGELOG.md
```diff
@@ -2,11 +2,16 @@
...
+## [unreleased]
+
+### Added
+
+- `agate/TabGroup` prop `onTabClick` to handle `onClick` event on it
 
 ### Fixed
 
 - `agate/Panels/TabbedPanels` to not show console error when there is no children
+- `agate/TabGroup` to pass onSelect to `ui/Group`
```
stanca-pop-lgp, line 14: Please add backticks to onSelect as well.

stanca-pop-lgp (changes requested):
Please check out my comments.

stanca-pop-lgp (approved):
LGTM

samples/sampler/stories/default/TabGroup.js
```diff
@@ -13,8 +14,10 @@ export default {
 
 export const _TabGroup = (args) => {
 	const orientation = args['orientation'];
+
 	return (
 		<TabGroup
+			onSelect={action('onSelect')}
```
jeonghee27, line 20: The qa-a11y sample also has a TabGroup. Could you please fix qa-a11y sampler also?

TabGroup/TabGroup.js
```diff
@@ -73,7 +74,7 @@ const TabBase = kind({
 		 * @type {Function}
 		 * @public
 		 */
-		onClick: PropTypes.func,
+		onTabClick: PropTypes.func,
```
jeonghee27, line 77: Can't we keep the onClick prop name as it is?

TabGroup/TabGroup.js
```diff
@@ -212,6 +221,18 @@ const TabGroupBase = kind({
...
+		 * Called when an item is selected.
+		 *
+		 * The event payload will be an object with the following members:
+		 * * `data` - The value for the option as received in the `children` prop
+		 * * `selected` - Number representing the selected option, 0 indexed
+		 *
+		 * @type {Function}
+		 * @public
+		 */
+		onSelect: PropTypes.func,
```
jeonghee27, line 234: It also need changelog

CHANGELOG.md
```diff
@@ -2,11 +2,16 @@
 
 The following is a curated list of changes in the Enact agate module, newest changes on the top.
 
-## Unreleased
+## [unreleased]
+
+### Added
+
+- `agate/TabGroup` prop `onTabClick` to handle `onClick` event on it
```
jeonghee27, line 9: onTabClick is not a prop for TabGroup.
it's prop for TabBase. but TabBase is private

TabGroup/TabGroup.js
jeonghee27: Please leave the "cause" of why "onClick" does not work.

TabGroup/tests/TabGroup-specs.js
```diff
@@ -1,11 +1,30 @@
...
 import ThemeDecorator from '../../ThemeDecorator';
 
 describe('TabGroup Specs', () => {
+	test('should fire `onTabClick` with `onTabClick` type when a tab is clicked', () => {
+		const handleTabClick = jest.fn();
+		render(
+			<TabGroup
+				tabPosition="before"
+				tabs={[
+					{title: 'Home', icon: 'home', onTabClick: handleTabClick},
```
jeonghee27, line 16: When I tried with "selected:true"

{title: 'Home', icon: 'home', selected:true}, It does not work neither.
Not all props we put here will be passed to TabBase. (I don't know whether it is a bug or original intention)

I guess the method of putting onClick here wasn't supported from the beginning. (Because TabGroup support onSelect)
I wonder if it is correct to fix about onClick.

samples/sampler/stories/default/TabGroup.js
jeonghee27: Can you add the "selectedIndex" prop to make show the selected state? The selection style is clear on the Carbon skin.

TabGroup/tests/TabGroup-specs.js
andreitirlalge: Whenever I tried this fix with onClick prop, inside GroupItem.js(@enact/ui) it's being overwritten into handleSelect this resulting into a bug because onClick is not being passed further to the component. 
image
What do you suggest?

TabGroup/TabGroup.js
andreitirlalge: Can you check my comment below?

TabGroup/tests/TabGroup-specs.js
jeonghee27: > I guess the method of putting onClick here wasn't supported from the beginning. (Because TabGroup support onSelect)
I wonder if it is correct to fix about onClick.

I think the onClick included in each item may not have been intentionally disconnected.
Is there any problem if this is not supported?
Supporting onClick doesn't seem to have any documentation or examples.

samples/qa-a11y/src/views/TabGroup.js
```diff
@@ -10,18 +10,29 @@ const tabs = [
...
+	const [selectedAriaLabelled, setSelectedAriaLabelled] = useState(0);
+
+	const handleSelected = useCallback(({selected}) => setSelectedNormal(selected), []);
+	const handleSelectedAriaLabelled = useCallback(({selected}) => setSelectedAriaLabelled(selected), []);
+
+	return (
+		<>
+			<Section title="Default">
+				<TabGroup alt="Normal" onSelect={handleSelected} selectedIndex={selectedNormal} tabPosition="before" tabs={tabs} />
+				<TabGroup alt="Disabled" disabled tabPosition="before" tabs={tabs} />
```
jeonghee27, line 24: I'll Add onSelect handler also (To check disabled functionality)

jeonghee27 (approved):
LGTM

PR #716, opened 2023-01-13
WRP-6584: Fixed failed ui tests on Jenkins

stanca-pop-lgp:

### Checklist

Ui tests job on Jenkins for Agate develop branch are failing for Drawer (with scrimType - none should have correct heading) and SwitchItem (pointer should re-select the item when clicked twice).

### Resolution
For Drawer I changed `waitForOpen` and `waitForClose` to be asyncronous. They both use `waitForExist` command from webdriverio which must be used inside async functions (https://webdriver.io/docs/api/element/waitForExist/).
For SwitchItem I refocused the element before selecting it again.

### Additional Considerations
When running the changes a new test failed for TimePicker so I added a browser pause to wait for the animation to be completed.

### Links
WRP-6584

### DIFF
--- a/tests/ui/specs/SwitchItem/SwitchItem-specs.js
+++ b/tests/ui/specs/SwitchItem/SwitchItem-specs.js
@@ -79,7 +79,6 @@ describe('SwitchItem', function () {
 			it('should re-select the item when selected twice', async function () {
 				await switchItem.focus();
 				await Page.spotlightSelect();
-				await switchItem.focus();
 				await Page.spotlightSelect();
 				expect(await switchItem.isSelected).to.be.true();
 			});
--- a/tests/ui/specs/SwitchItem/SwitchItem-specs.js
+++ b/tests/ui/specs/SwitchItem/SwitchItem-specs.js
@@ -79,6 +79,7 @@ describe('SwitchItem', function () {
 			it('should re-select the item when selected twice', async function () {
 				await switchItem.focus();
 				await Page.spotlightSelect();
+				await switchItem.focus();
 				await Page.spotlightSelect();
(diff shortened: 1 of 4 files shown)


daniel-stoian-lgp (approved):

adrian-cocoara-lgp:
The solution for adding `async await` for `waitForExist` function comes from version 8 of wdio documentation.
We are using v7 in `ui-test-utils`, and the docs for v7 does not include `async await`:
https://v7.webdriver.io/docs/api/element/waitForExist

We need to find another solution

adrian-cocoara-lgp (approved):

tests/ui/specs/SwitchItem/SwitchItem-specs.js
```diff
@@ -79,6 +79,7 @@ describe('SwitchItem', function () {
 			it('should re-select the item when selected twice', async function () {
 				await switchItem.focus();
 				await Page.spotlightSelect();
+				await switchItem.focus();
```
dongsuwon, line 82: I'm not sure that adding this step will solve the failed test. Failure did not occur after this issue was reported.
Is there a reason to refocus? Please let me know.

dongsuwon (changes requested):

tests/ui/specs/SwitchItem/SwitchItem-specs.js
stanca-pop-lgp: I applied by mistake changes to the wrong test. I thought the one with 5-way fails and that it loses focus. I will check the pointer test and come up with a fix if it still fails. Thank you for noticing!

dongsuwon (approved):

dongsuwon:
LGTM

PR #727, opened 2023-02-23
WRP- 8541: Added unit tests for data/ConsumerDecorator

adrian-cocoara-lgp:

### Checklist

Added unit tests for `data/ConsumerDecorator`

### Resolution

### Additional Considerations
Fixed typo for `componentWillUnmount` keyword in data/ProviderDecorator

### Links
WRP-8541


data/ConsumerDecorator/tests/ConsumerDecorator-specs.js
```diff
@@ -0,0 +1,99 @@
...
+// ComponentB created for testing `mapStateToProps` defined as an object
+const BaseComponentB = (props) => {
+	componentBProps = props;
+
+	return <AppContext.Provider value="test"><div /></AppContext.Provider>;
+};
+
+const ComponentBDecorator = compose(
+	ProviderDecorator({
+		pauseOnBlur: true,
```
juwonjeong, line 51: Do we need `pauseOnBlur` for this test case?

adrian-cocoara-lgp: Yes, we do.
This line was added for code coverage purposes. This prop covers lines 37-38 andn 44-45 for ProviderDecorator.js
Removing this line wou;d reduce code coverage to 60%:
image

juwonjeong: I'm afraid I cannot agree that pauseOnBlur is related to this test case. It can drop the code coverage of ProvideDecorator, but this test case is for testing ConsumerDecorator logic. And I think this TC is not designed for pauseOnBlur: true case(line 37-38).

I think to keep pauseOnBlur: true, we need to include test code like the below for example.
`expect(window.addEventListener).toBeCalledWith('blur', expect.any(Function));`

adrian-cocoara-lgp: I tried writing a TC for `pauseOnBlur` prop, but I can't find a way to access it.
So I think we can go ahead with your suggestion and remove it. Thank you!

juwonjeong (approved):
LGTM

PR #751, opened 2023-07-12
WRP-21123: Added missing prop from sampler and applied the required css

andreitirlalge:
### Checklist

Added missing prop `cover` from storybook.
Applied the required css for the prop value `partial`.

### Resolution
Panels now displays the prop `cover` on storybook with the required css.

### Additional Considerations

### Links
WRP-21123

### Comments
Enact-DCO-1.0-Signed-off-by: Andrei Tirla andrei.tirla@lgepartner.com

### DIFF
--- a/styles/variables-carbon.less
+++ b/styles/variables-carbon.less
@@ -44,7 +44,7 @@
 @agate-panels-controls-top: (@agate-component-spacing - 6px);
 @agate-panels-panel-padding: 24px;
 @agate-panels-partial-height: 50%;
-@agate-panels-partial-width: auto;
+@agate-panels-partial-width: 890px;
 @agate-panels-tab-margin: 0;
 @agate-panels-tab-padding: 18px 0;
 @agate-panels-tab-separator-border-width: 1px;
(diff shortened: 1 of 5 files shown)


ion-andrusciac-lgp (approved):

styles/variables-carbon.less
```diff
@@ -44,7 +44,7 @@
 @agate-panels-controls-top: (@agate-component-spacing - 6px);
 @agate-panels-panel-padding: 24px;
 @agate-panels-partial-height: 50%;
-@agate-panels-partial-width: auto;
+@agate-panels-partial-width: 890px;
```
sjro, line 47: Why did you write a specific pixel number? Not 50%?

andreitirlalge: I wrote a specific pixel number because I tried to maintain the same pattern we used in the other styles. e.g (styles/variables-gallium.less). Do you think it would be better to set it to 50%?

sjro: No. I will merge this PR.

PR #757, opened 2023-09-07
WRP-26873: Fixed partial Panels' transition on horizontal

ion-andrusciac-lgp:
### Checklist

Fixed Panels partial transition on horizontal

### Resolution
Changed component styling that will allow to apply the correct CSS properties on component transition

### Additional Considerations

### Links
WRP-26873

### Comments

Enact-DCO-1.0-Signed-off-by: Ion Andrusciac ion.andrusciac@lgepartner.com

### DIFF
--- a/Panels/Panels.module.less
+++ b/Panels/Panels.module.less
@@ -79,6 +79,7 @@
 			background-color: @agate-panels-partial-bg-color;
 			background-image: @agate-panels-partial-bg-image;
 			height: @agate-panels-partial-height;
+			overflow: hidden;
 			width: @agate-panels-partial-width;
 		}
 
--- a/Panels/Panels.module.less
+++ b/Panels/Panels.module.less
@@ -6,7 +6,7 @@
 
 .panels {
 	&,
-	& .viewport {
+	.viewport {
 		position: absolute;
 		.position(0);
 	}
@@ -62,6 +62,13 @@
 			.remove-margin-on-edge-children();
 		}
 
+		.viewport {
+			&.transitioning {
+				-webkit-clip-path: polygon(0 0, 100% 0, 100% 100%, 0 100%);
+				clip-path: polygon(0 0, 100% 0, 100% 100%, 0 100%);
+			}
+		}
+
 		&.partial {
 			background-color: @agate-panels-partial-bg-color;
 			background-image: @agate-panels-partial-bg-image;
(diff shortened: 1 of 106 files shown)


Panels/Panels.module.less
```diff
@@ -82,17 +89,6 @@
...
-			.viewport {
-				&.transitioning {
-					-webkit-clip-path: polygon(@agate-panels-breadcrumb-width 0, 100% 0, 100% 100%, @agate-panels-breadcrumb-width 100%);
-					clip-path: polygon(@agate-panels-breadcrumb-width 0, 100% 0, 100% 100%, @agate-panels-breadcrumb-width 100%);
-				}
-
-				> :not([data-index="0"]) {
-					width: calc(~"100% - " @agate-panels-breadcrumb-width);
-				}
-			}
```
daniel-stoian-lgp, line 95: is this removal of styling for breadcrumbpanels needed?

daniel-stoian-lgp:
Can you add a changelog? Panels is a public component

Panels/Panels.module.less
ion-andrusciac-lgp: yes, because styling from here is not applied to the `Viewport` component

daniel-stoian-lgp (approved):

Panels/Panels.module.less
```diff
@@ -69,6 +69,13 @@
 			}
 		}
 
+		.viewport {
+			&.transitioning {
+				-webkit-clip-path: polygon(0 0, 100% 0, 100% 100%, 0 100%);
```
MikyungKim, line 74: According to MDN, we don't need to add `-webkit-` prefix for `clip-path`.

Panels/Viewport.js
```diff
@@ -138,7 +138,9 @@ const ViewportBase = class extends Component {
 
 	removeTransitioningClass = () => {
 		if (this.node) {
-			this.node.classList.remove(css.transitioning);
+			setTimeout(() => {
```
MikyungKim, line 141: I saw the flickering but could you explain why?
🤔 Is there any other way instead of `setTimeout`?

ion-andrusciac-lgp: I checked again and I did not see the flickering, and `setTimeout` was used to prevent it. Can you please verify again or help me reproduce this behavior?

MikyungKim: I meant, if I removed `setTimeout`, I can see the flickering.
What if we remove `setTimeout` here and add `overflow: hidden` property in the `.panels.partial` from `Panels.module.less`?
I think we don't need to draw outside of Panels if it covers partial.
What do you think? Do you have any conerns?

ion-andrusciac-lgp: I checked, and it looks good. Thank you!

Panels/Panels.module.less
```diff
@@ -69,10 +69,17 @@
...
+				clip-path: polygon(0 0, 100% 0, 100% 100%, 0 100%);
+			}
+		}
+
 		&.partial {
 			background-color: @agate-panels-partial-bg-color;
 			background-image: @agate-panels-partial-bg-image;
 			height: @agate-panels-partial-height;
+			overflow: hidden;
 			width: @agate-panels-partial-width;
```
MikyungKim, line 83:
```suggestion
			width: @agate-panels-partial-width;
			overflow: hidden;
```

Panels/tests/Panels-specs.js
```diff
@@ -13,6 +13,17 @@ const escapeKeyUp = keyUp(27);
 const escapeKeyDown = keyDown(27);
 
 describe('Panels Specs', () => {
+	beforeEach(() => {
```
MikyungKim, line 16: Could you also remove this change?

MikyungKim (approved):
LGTM

PR #854, opened 2026-01-29
NXT-8995: Adapt Agate code to respect React Compiler

alexandrumorariu:
### Checklist

* [X] I have read and understand the [contribution guide](http://enactjs.com/docs/developer-guide/contributing/)
* [X] Documentation was added or is not needed

Adapted Agate code to respect React Compiler

### Resolution
eslint-plugin-react-hooks 7.x has React Compiler embedded. The code needs to be adapted to respect Eslint rules

- Adapted how the refs were changed during render, and also cannot read .current during render. This is not allowed outside handlers or effects [REFERENCE](https://react.dev/reference/react/useRef#:~:text=You%20can%20read%20or%20write%20refs%20from%20event%20handlers%20or%20effects%20instead.)
- Refs.current cannot be used in dependency array
- Modifying a variable defined outside a component or hook is not allowed
- Modifying component props or hook arguments is not allowed
- State update inside useEffect is not recommended

### Additional Considerations
Some of the errors were silenced as they were considered false positives (future investigation may be required)

### Links
NXT-8995


useScroll/ScrollButtons.js
```diff
@@ -201,6 +159,48 @@ const useScrollButtons = (props) => {
...
+		const prevRef = prevButtonRef.current;
+
+		utilEvent('keydown').addEventListener(nextRef, onKeyDownNext);
+		utilEvent('keydown').addEventListener(prevRef, onKeyDownPrev);
+
+		return () => {
+			utilEvent('keydown').removeEventListener(nextRef, onKeyDownNext);
+			utilEvent('keydown').removeEventListener(prevRef, onKeyDownPrev);
+		};
+	});
```
daniel-stoian-lgp, line 173: maybe we can check in a future ticket why this useEffect had no dependencies

daniel-stoian-lgp (approved):
LGTM

Scroller/useThemeScroller.js
```diff
@@ -215,7 +215,7 @@ const useSpottable = (props, instances) => {
 		const itemRect = getRect(item);
 
 		if (horizontal && !(itemRect.left >= containerRect.left && itemRect.right <= containerRect.right)) {
-			scrollContentHandle.current.scrollPos.left = calculateScrollLeft(item, scrollPosition);
+			scrollContentHandle.current.scrollPos.left = calculateScrollLeft(item, scrollPosition); // eslint-disable-line react-hooks/immutability
```
ion-andrusciac-lgp: I think instead of disabling ESLint, we can define a local variable like `const scrollContentHandleRef = scrollContentHandle.current;` and use this const where `scrollContentHandle.current` is used

useScroll/Scrollbar.js
```diff
@@ -97,7 +97,7 @@ const ScrollbarBase = memo(({css = componentCss, minThumbSize = 18, vertical = t
 		const {scrollbarHandle} = props;
 		const {update: uiUpdate} = scrollbarHandle.current;
 
-		scrollbarHandle.current.update = (bounds) => {
+		scrollbarHandle.current.update = (bounds) => {  // eslint-disable-line react-hooks/immutability
```
ion-andrusciac-lgp: We can rename `scrollbarHandle` to `scrollbarHandleRef`, and in this way, ESLint will not throw an error because references can be modified

VirtualList/useSpotlight.js
```diff
@@ -124,7 +124,7 @@ const useSpotlightRestore = (props, instances, context) => {
 				mutableRef.current.restoreLastFocused = false;
 
 				// try to focus the last focused item
-				spottable.current.isScrolledByJump = true;
+				spottable.current.isScrolledByJump = true; // eslint-disable-line react-hooks/immutability
```
ion-andrusciac-lgp: And here we can declare a local variable

useScroll/useScroll.js
```diff
@@ -192,7 +192,7 @@ const useThemeScroll = (props, instances) => {
 		// oddly, Scroller manages scrollContainerHandle.current.bounds so if we don't update it here (it is also
 		// updated in calculateAndScrollTo, but we might not have made it to that point), it will be
 		// out of date when we land back in this method next time.
-		scrollContainerHandle.current.bounds.scrollHeight = scrollContainerHandle.current.getScrollBounds().scrollHeight;
+		scrollContainerHandle.current.bounds.scrollHeight = scrollContainerHandle.current.getScrollBounds().scrollHeight;  // eslint-disable-line react-hooks/immutability
```
ion-andrusciac-lgp: Same here

ion-andrusciac-lgp (approved):

PR #868, opened 2026-07-03
NXT-13025: enabled 2 addons for storybook

daniel-stoian-lgp:
### Checklist

* [x] I have read and understand the [contribution guide](https://enactjs.com/docs/developer-guide/contributing/)
* [ ] A [CHANGELOG entry](https://enactjs.com/docs/developer-guide/contributing/changelogs/) is included

Adopts 2 Storybook addons in the Limestone sampler: pseudo-states and viewport.

### Resolution
1. Pseudo-states (storybook-addon-pseudo-states)

Added the dependency and registered the addon so :hover / :focus / :active etc. can be toggled from the toolbar on any story.
Set parameters.pseudo.rootSelector = 'body' globally: Enact's focus rules are scoped under .spotlight-input-* on #storybook-root, so the addon's forcing class must sit on a strict ancestor of that element (the default #storybook-root root doesn't work).
Added a PseudoStateFocusBridge decorator. The addon only flips CSS classes, but Enact focus visuals are gated by Spotlight input-mode and, for components like TooltipDecorator, by React state. The bridge observes the addon's forced state and mirrors it with real Enact focus (.focus() + a dispatched focusin + spotlight-input-key), so focus styling and state-driven UI (e.g. tooltips) actually respond.

2. Viewport (Storybook 10 core feature)
Enabled via features.viewport 
Added TV and portrait presets via parameters.viewport.options: 720p / 1080p / 4K landscape and 1080p / 4K portrait.
Added manager-head.html CSS so that when a viewport larger than the preview area is selected, both scrollbars stay pinned to the visible edges 

### Additional Considerations

### Links
NXT-13025

### Comments
Enact-DCO-1.0-Signed-off-by: Daniel Stoian ([daniel.stoian@lgepartner.com](mailto:daniel.stoian@lgepartner.com))

### DIFF
--- a/samples/sampler/.qa-storybook/main.js
+++ b/samples/sampler/.qa-storybook/main.js
@@ -18,7 +18,7 @@ export default {
 		backgrounds: false,
 		interactions: false,
 		postcss: false,
-		viewport: false,
+		viewport: true,
 		warnOnLegacyHierarchySeparator: false
 	},
 	framework: {
@@ -43,6 +43,7 @@ export default {
 	addons: [
 		'@enact/storybook-utils/addons/actions',
 		'@enact/storybook-utils/addons/controls',
+		'storybook-addon-pseudo-states',
 		...(process.env.PERF_PANEL === 'true' ? ['@github-ui/storybook-addon-performance-panel'] : [])
 	],
 	managerEntries: [resolve(__dirname, '../custom-addon/manager.js')],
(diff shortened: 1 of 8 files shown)


alexandrumorariu (approved):
LGTM

ion-andrusciac-lgp (approved):

PR #874, opened 2026-08-24
NXT-19371: resolve webpack version mismatch causing parser.getLocatio…

hong6316:
### Checklist

* [ ] I have run automated testing and it is passed

The build fails during the Storybook webpack build with:

```
Module parse failed: parser.getLocation is not a function
TypeError: parser.getLocation is not a function
  at constDependency (JavascriptParserHelpers.js:129:20)
  at DefinePlugin.js:951:10
```

Two conflicting webpack versions are installed simultaneously:
- **webpack@5.107.2** — pinned in `devDependencies` (the "host" webpack used by Storybook)
- **webpack@5.109.2** — installed nested under `@enact/dev-utils/node_modules/` (resolved from peer dependencies)

When `@enact/dev-utils`'s `ILibPlugin` does `require('webpack')`, it resolves to the nested **5.109.2** copy. The `DefinePlugin` from 5.109.2 registers parser hooks that call `parser.getLocation()` — a method introduced in **webpack 5.108+**. However, the actual parser instance comes from the host's webpack 5.107.2, which **does not have this method**, causing the `TypeError`.

### Resolution
1. **Update storybook** from `^10.4.6` to `^10.5.10` to align with the version npm resolves for nested dependencies

### Additional Considerations

### Links
NXT-19371

### Comments
Enact-DCO-1.0-Signed-off-by: Taeyoung Hong (taeyoung.hong@lge.com)


samples/sampler/patches/storybook+10.5.10.patch
```diff
@@ -1,22 +0,0 @@
-diff --git a/node_modules/storybook/dist/manager/globals-runtime.js b/node_modules/storybook/dist/manager/globals-runtime.js
```
daniel-stoian-lgp, line 1: same here. why was this removal needed?

hong6316: I had deleted this patch because it caused a build error.

Upon re-verification, I found that the patch was incompatible with the Storybook version, so I recreated it as v10.5.10.

daniel-stoian-lgp (approved):

